PR #81, opened 2023-05-08
Fixes "too many clients" issue from PG by setting max open connections

snormore:
When the syncer gets backed up significantly, a bunch of pg connections get opened, resulting in "too many clients" errors. This PR mitigates that by configuring [MaxOpenConns](https://pkg.go.dev/database/sql#DB.SetMaxOpenConns).


snormore:
I added an issue for tracking the connection pooling work: https://github.com/orgs/xmtp/projects/8/views/5

Can I get a 🍏 on this one 🙏

Issue #56, opened 2023-04-20
Performance and reliability evaluation


m-005:
https://www.notion.so/XMTP-Network-Performance-and-Reliability-Evaluation-Plan-ea294cbf2e2741c4b98131331357574b

PR #83, opened 2023-05-09
e2e: convergence test

snormore:
This PR replaces the e2e `publish subscribe query` test with the `convergence` test, which is similar except does not immediately fail if a node isn't available or if unable to converge, instead reporting status of convergence per client/node with durations. The intention of this is to give visibility into consistency latency across nodes.

https://github.com/xmtp/xmtpd/issues/56

Screenshot 2023-05-08 at 9 04 10 PM


pkg/e2e/test_convergence.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+func isErrClosedConnection(err error) bool {
+	return errors.Is(err, io.EOF) || strings.Contains(err.Error(), "closed network connection") || strings.Contains(err.Error(), "response body closed")
+}
+
+func expectQueryMessagesEventually(ctx context.Context, client apiclient.Client, contentTopics []string, expectedEnvs []*messagev1.Envelope) error {
+	timeout := 10 * time.Second
+	delay := 10 * time.Millisecond
+	// TODO: higher delay for production?
```
snormore, line 235: Lower delay gives more precision on the convergence durations, but adds more query load. To be addressed before merging this PR.

neekolas (approved):

PR #92, opened 2024-07-26
Start from empty base

richardhuaaa:
Doing this in pieces for clarity. This moves the old implementation to a [historical](https://github.com/xmtp/xmtpd/tree/historical) branch and gives us a clean starting point to work from. I kept the license and README and tweaked it slightly.


neekolas (approved):
So much red

PR #94, opened 2024-07-27
Copy over server scaffolding

richardhuaaa:
Basically moves over the work from https://github.com/xmtp/xmtp-node-go/pull/397, as well as some utils from the old repo and a clean go dependency list.

- Sets up basic server
- Sets up basic API server using generated protos
- Sets up DB
- Sets up interface for `NodeRegistry`

Still need to copy over the CI and docker


go.mod
```diff
@@ -1,3 +1,64 @@
 module github.com/xmtp/xmtpd
 
 go 1.21.11
```
neekolas, line 3: We should switch to go 1.22, since we are starting fresh and that's the current stable version

replication/main.go
```diff
@@ -0,0 +1,116 @@
+package main
```
neekolas, line 1: It's convention in Go packages to put these in a `cmd` folder. Think we should keep that convention

neekolas:
Oof. I missed that this PR exists and made an [almost identical version](https://app.graphite.dev/github/pr/xmtp/xmtpd/95/Bootstrap-service).

Think the only real difference is that I left out the DB stuff, since I'd like to take a fresh look at the DB migrations since we don't need Bun anywhere in this repo

PR #102, opened 2024-08-09
chain streaming interfaces

neekolas:
## tl;dr

- Adds interfaces for streaming raw log messages from smart contracts on a blockchain
- Adds configuration for building and configuring the log streamer
- Mocks the Geth client for chain streaming

## Notes

There are many ways to get messages from a blockchain, each with trade-offs.

1. Query `eth_getLogs` on an archive node to retrieve log messages
2. Query `eth_getLogs` on an archive node to retrieve old log messages, then subscribe for new log messages in real-time
3. Have an ethereum node pipe log messages to stdout and have them ingested directly into a better archival service.

I have opted for 1 here, since it's the simplest and most reliable. We can move to 2 if we want to improve latency for new blocks.

Moving to 3 is more complicated. StreamingFast has a [Firehose](https://firehose.streamingfast.io/) package that can be used on top of a node to rapidly ingest logs, and deal with complicated cases like reorgs. But that takes a bunch of infrastructure that I'm not sure node operators are going to be up to. Options 1 and 2 work with any blockchain node.

It may be possible to work with StreamingFast to get a hosted Firehose deployed that node operators can work with, possibly via [Substreams](https://substreams.streamingfast.io/) to sink directly into a DB. But that's going to take time.


neekolas:
* **#102** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/102?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pkg/indexer/blockchain/rpcLogStreamer.go
```diff
@@ -0,0 +1,167 @@
...
+		return nil, err
+	}
+	return NewRpcLogStreamer(client, c.logger, c.contractConfigs), nil
+}
+
+// Struct defining all the information required to filter events from logs
+type contractConfig struct {
+	fromBlock       int
+	contractAddress common.Address
+	topics          []common.Hash
```
richardhuaaa, line 53: Are these topics the same as the topics that users would query by? Is this included so that users can query the blockchain directly? For nodes, I imagine they would just firehose everything

neekolas: No, "topics" is an ethereum event term as well. In this case, the topic is actually the hash of the signature of the event itself so that the indexer can pull the right events in contracts that might emit multiple event types.

pkg/indexer/blockchain/rpcLogStreamer.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (r *RpcLogStreamer) getNextPage(config contractConfig, fromBlock int) (logs []types.Log, nextBlock *int, err error) {
+	highestBlock, err := r.client.BlockNumber(r.ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	highestBlockCanProcess := int(highestBlock) - LAG_FROM_HIGHEST_BLOCK
+	numOfBlocksToProcess := highestBlockCanProcess - fromBlock + 1
```
richardhuaaa, line 123: Why is it +1 here? Is the fromBlock already processed, or not processed yet?

pkg/indexer/blockchain/rpcLogStreamer.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	// TODO:(nm) Use some more clever tactics to fetch the maximum number of logs at one times by parsing error messages
+	// See: https://github.com/joshstevens19/rindexer/blob/master/core/src/indexer/fetch_logs.rs#L504
+	logs, err = r.client.FilterLogs(r.ctx, buildFilterQuery(config, int64(fromBlock), int64(to)))
+	if err != nil {
+		return nil, nil, err
+	}
+
+	nextBlockNumber := to + 1
```
richardhuaaa, line 142: I see, so it looks like the `fromBlock` is not processed? I wonder if it would be easier to understand/math if we use the cursor pattern and make this `lastProcessedBlock` instead of `fromBlock`?

pkg/indexer/blockchain/rpcLogStreamer.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	return nil
+}
+
+func buildFilterQuery(contractConfig contractConfig, fromBlock int64, toBlock int64) ethereum.FilterQuery {
+	addresses := []common.Address{contractConfig.contractAddress}
+	topics := [][]common.Hash{}
+	for _, topic := range contractConfig.topics {
+		topics = append(topics, []common.Hash{topic})
+	}
```
richardhuaaa, line 159: Noob question, why is this a 2D array? What does each dimension represent?

richardhuaaa (approved):

pkg/indexer/blockchain/rpcLogStreamer.go
```diff
@@ -0,0 +1,167 @@
...
+	for _, watcher := range r.watchers {
+		go r.watchContract(watcher)
+	}
+	return nil
+}
+
+func (r *RpcLogStreamer) watchContract(watcher contractConfig) {
+	fromBlock := int(watcher.fromBlock)
+	logger := r.logger.With(zap.String("contractAddress", watcher.contractAddress.Hex()))
+	for {
```
richardhuaaa, line 92: This is an infinite poll loop without any interval between polls right? Curious if there's a rationale for having no interval?

pkg/indexer/blockchain/rpcLogStreamer.go
neekolas: It's the number of blocks you want to have in a single query. So even if you have `from: 5, to: 5` you are actually processing 1 block since it seems both the from and to are inclusive.

pkg/indexer/blockchain/rpcLogStreamer.go
neekolas: Great question. Here's the docs around it. 
https://pkg.go.dev/github.com/ethereum/go-ethereum#FilterQuery

pkg/indexer/blockchain/rpcLogStreamer.go
neekolas: The `fromBlock` should get processed here. Both from and to are inclusive in ethereum

pkg/indexer/blockchain/rpcLogStreamer.go
neekolas: Good call out. I can make it sleep once it gets to the point of having no new logs

neekolas:
### Merge activity

* **Aug 12, 2:33 PM PDT**: @neekolas started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/xmtp/xmtpd/102).

PR #105, opened 2024-08-10
Scaffold smart contract event storage

neekolas:
## tl;dr

- Adds function to run log streamer and store events
- Creates flexible interface for deserializing and storing events
- Updates mockery config and moves all mocks to a dedicated folder

## AI Assisted summary
- **blockchain/interfaces**: Remove `Stop()` method from `LogStreamer` interface, add custom logger name to `RpcLogStreamer`.
- **blockchain/rpcLogStreamer.go**: Add buffered channel to `ListenForContractEvent` to improve event handling.
- **indexer/indexer.go**: Add new file for starting the indexer service with configuration for message contract events.
- **storer**: Implement error handling for the log storer, including support for retryable errors. Add mock for `LogStorer`.
- **Refactoring**: Move existing mock generation to the 'mocks' package for better code organization.
- **server/options.go**: Integrate `ContractsOptions` into the server options for configuration.

PR #103, opened 2024-08-09
Add contracts

neekolas:
## tl;dr

- Adds smart contracts to repo
- Sets up Foundry in the repo 
- Adds tests for smart contracts
- Sets up `abigen` for generating code out of smart contracts

## Warning

You will need to run `./dev/up` to generate some code and start the blockchain node

You may also have to run `git submodule update --init --recursive` to pull down the submodules.

PR #106, opened 2024-08-10
Update node contract

neekolas:
### TL;DR

- Rewrites Nodes contract using the ERC721 standard as a base
- Adds a bunch of tests for the new smart contract
- Removes a deprecated flag from our mockery config

## Why ERC721

1. It doesn't seem crazy to model this registry as a NFT, so we might as well use the standard.
2. We get all the OpenZeppelin audited smart contract libraries for free
3. It'll work with the entire rest of the crypto ecosystem (external indexers, other smart contracts, etc) automatically

### AI Assisted Changelog

1. **Smart Contract Changes**:
   - Implemented ERC721 standard in the `Nodes` contract.
   - Added new functions for node management using NFTs: `addNode`, `removeNode`, `updateHttpAddress`, `updateMtlsCert`, and `updateHealth`.
   - Added events `NodeUpdated` and `NodeRemoved` for better tracking of node changes.
   - Restricted node ownership transfer.

2. **Go Bindings Update**:
   - Updated the ABI bindings to reflect new changes in the `Nodes` contract.
   - Added data structures `NodesNode` and `NodesNodeWithId` to handle the new Node data.
   - Implemented necessary function bindings for node management in Go.

3. **Testing**:
   - Added new test cases in Solidity to cover the new node management functionalities.
   - Renamed test contract from `CounterTest` to `GroupMessagesTest` and added `NodesTest` for comprehensive testing of node functionalities.

### DIFF
--- a/contracts/src/Nodes.sol
+++ b/contracts/src/Nodes.sol
@@ -1,91 +1,162 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.13;
 
+import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
 
-contract Nodes is Ownable {
-    constructor() Ownable(msg.sender) {}
+/**
+A NFT contract for XMTP Node Operators.
 
+The deployer of this contract is responsible for minting NFTs and assinging them to node operators.
+
+All nodes on the network periodically check this contract to determine which nodes they should connect to.
+ */
+contract Nodes is ERC721, Ownable {
+    constructor() ERC721("XMTP Node Operator", "XMTP") Ownable(msg.sender) {}
+
+    // uint16 counter so that we cannot create more than 65k IDs
+    // The ERC721 standard expects the tokenID to be uint256 for standard methods unfortunately
+    uint16 private _nodeIdCounter;
+
+    // A node, as stored in the internal mapping
     struct Node {
+        bytes signingKeyPub;
         string httpAddress;
-        uint256 originatorId;
         bool isHealthy;
-        // Maybe we want a TLS cert separate from the public key for MTLS authenticated connections?
     }
 
-    event NodeUpdate(
-        bytes publicKey,
-        string httpAddress,
-        uint256 originatorId,
-        bool isHealthy
-    );
+    struct NodeWithId {
+        uint16 nodeId;
+        Node node;
+    }
 
-    // List of public keys
-    bytes[] publicKeys;
+    event NodeUpdated(uint256 nodeId, Node node);
 
-    // Mapping of publicKey to node
-    mapping(bytes => Node) public nodes;
+    // Mapping of token ID to Node
+    mapping(uint256 => Node) private _nodes;
 
     /**
-    Add a node to the network
+    Mint a new node NFT and store the metadata in the smart contract
      */
     function addNode(
-        bytes calldata publicKey,
+        address to,
+        bytes calldata signingKeyPub,
         string calldata httpAddress
-    ) public onlyOwner {
-        require(
-            bytes(nodes[publicKey].httpAddress).length == 0,
-            "Node already exists"
-        );
+    ) public onlyOwner returns (uint16) {
+        uint16 nodeId = _nodeIdCounter;
+        _mint(to, nodeId);
+        _nodes[nodeId] = Node(signingKeyPub, httpAddress, true);
+        _emitNodeUpdate(nodeId);
+        _nodeIdCounter++;
 
-        require(bytes(httpAddress).length != 0, "HTTP address is required");
-
-        nodes[publicKey] = Node({
-            httpAddress: httpAddress,
-            originatorId: publicKeys.length + 1,
-            isHealthy: true
-        });
-
-        publicKeys.push(publicKey);
-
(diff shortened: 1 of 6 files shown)


neekolas:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/106?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#106** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/106?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#105** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/105?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#103** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/103?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#102** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/102?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

contracts/test/Nodes.sol
```diff
@@ -0,0 +1,159 @@
...
+        Nodes.Node memory node = _randomNode(true);
+        address operator = vm.randomAddress();
+
+        uint16 nodeId = nodes.addNode(
+            operator,
+            node.signingKeyPub,
+            node.httpAddress
+        );
+
+        vm.prank(operator);
```
neekolas, line 122: `vm.prank` is cool. It just lets you change the `msg.sender` value to any other address for a test, so you can impersonate someone and see how the contract behaves.

pkg/registry/registry.go
```diff
@@ -1,11 +1,10 @@
...
 type Node struct {
-	Index         int
-	PublicKey     []byte
-	GrpcAddress   string
-	DisabledBlock *uint64
-	// Maybe add mTLS cert here
+	NodeId      int
+	SigningKey  []byte
+	HttpAddress string
+	MtlsCert    []byte
```
neekolas, line 7: This field goes away in the next PR

contracts/src/Nodes.sol
```diff
@@ -1,91 +1,162 @@
...
-    Add a node to the network
+    Mint a new node NFT and store the metadata in the smart contract
      */
     function addNode(
-        bytes calldata publicKey,
+        address to,
+        bytes calldata signingKeyPub,
+        string calldata httpAddress
+    ) public onlyOwner returns (uint16) {
+        uint16 nodeId = _nodeIdCounter;
```
richardhuaaa, line 46: We should not allow a node to have node ID 0, because an originator SID of 0 means that the originator is the blockchain. Wondering if we should pre-increment _nodeIdCounter rather than post-incrementing it?

contracts/src/Nodes.sol
```diff
@@ -1,91 +1,162 @@
...
+     */
+    function transferFrom(
+        address from,
+        address to,
+        uint256 tokenId
+    ) public override {
+        require(
+            _msgSender() == owner(),
+            "Only the contract owner can transfer Node ownership"
+        );
```
richardhuaaa, line 69: Small nit, any reason this one uses a require statement, but `updateHealth` uses `public onlyOwner`?

contracts/src/Nodes.sol
```diff
@@ -1,91 +1,162 @@
...
+     */
+    function updateHttpAddress(
+        uint256 tokenId,
         string calldata httpAddress
-    ) public onlyOwner {
+    ) public {
         require(
-            bytes(nodes[publicKey].httpAddress).length == 0,
-            "Node already exists"
+            _msgSender() == ownerOf(tokenId),
```
richardhuaaa, line 81: I wonder if there are tradeoffs to allowing/encouraging nodes to update the HTTP address at any point, rather than going through a more heavyweight process via the DAO? For example, ensuring there is a graceful failover etc. Don't have a super strong opinion here

richardhuaaa (approved):
Using ERC721 is a smart idea. Will ecosystem tools break if we don't specify base/token URI? I suppose we could also generate some fun images too. Also do we need to block `safeTransferFrom`?

contracts/src/Nodes.sol
neekolas: I have a comment somewhere about this. I kinda think we should start at 1000 just to be safe. Would let us partition the smart contracts later if we wanted to.

contracts/src/Nodes.sol
neekolas: Only that I wanted a better error message, but agree the inconsistency feels weird

contracts/src/Nodes.sol
neekolas: I don't really know. Wondering the same thing. Worth a discussion.

neekolas (commented):
> Will ecosystem tools break if we don't specify base/token URI?

The OpenZeppelin default just sets it to "". I would expect tools to handle that, since some generative NFTs are on-chain only. But I'm with you that at some point we'll do an extra-credit project and make a URI that has JSON metadata with a fun image and more info. Then it fully looks like a normal NFT.

> Also do we need to block safeTransferFrom?

`safeTransferFrom` calls `transferFrom` internally, so we get them both with this override.

PR #107, opened 2024-08-11
Use new node contract in registry

neekolas:
## tl;dr

- Creates new `SmartContractRegistry` that polls the smart contract to get the latest state of the nodes
- Refactored the config out of the `server` to avoid circular dependencies and allow config objects to be passed in to other modules
- Added a `dev/run` command that populates all the environment variables required
- Created a `notifier` struct to power the `SmartContractRegistry` and publish changes to subscriptions

## Why not use the indexer?

We could get the same result from indexing the chain based on events. This seemed more straightforward. We don't care about historical state here, so processing old events is just going to create noise and complexity. When any node starts up it just needs a full picture of all the other nodes right now. We also don't have tight latency requirements: this contract moves in the cadence of days or weeks.

Since we know the node list is going to be small forever, and this data is going to be frequently accessed, having it all live in memory is also way more performant.


neekolas:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/107?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#107** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/107?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#106** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/106?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#105** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/105?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#103** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/103?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#102** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/102?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

README.md
```diff
@@ -8,9 +8,9 @@
 
 `xmtpd` (XMTP daemon) is an experimental version of XMTP node software. It is **not** the node software that currently forms the XMTP network.
 
-After `xmtpd` meets specific functional requirements, the plan is for it to become the node software that powers the XMTP network.
```
neekolas, line 11: Think my editor has a default Markdown formatter. We can make that standard issue in the repo so we don't get diffs like this.

pkg/registry/notifier_test.go
```diff
@@ -0,0 +1,85 @@
...
+	defer cancel()
+
+	for i := 0; i < 100; i++ {
+		go registry.trigger(1)
+	}
+	time.Sleep(30 * time.Millisecond)
+	require.Equal(t, 100, getCurrentCount())
+}
+
+func countChannel[Kind any](ch <-chan Kind) func() int {
```
neekolas, line 69: This was to prevent go's race detection from getting mad about data races in our tests

pkg/server/server.go
```diff
@@ -72,5 +74,5 @@ func (s *ReplicationServer) Shutdown() {
 }
 
 func parsePrivateKey(privateKeyString string) (*ecdsa.PrivateKey, error) {
-	return crypto.HexToECDSA(privateKeyString)
+	return crypto.HexToECDSA(strings.TrimPrefix(privateKeyString, "0x"))
```
neekolas, line 77: Lets us re-use the PK environment variable we are using with forge to deploy the contract locally, which expects a `0x` prefix. HexToECDSA does not expect the `0x`

pkg/registry/contractRegistry.go
```diff
@@ -0,0 +1,215 @@
...
+
+func (s *SmartContractRegistry) processNewNodes(nodes []Node) {
+	s.logger.Info("processing new nodes", zap.Int("count", len(nodes)), zap.Any("nodes", nodes))
+	s.newNodesNotifier.trigger(nodes)
+
+	s.nodesMutex.Lock()
+	defer s.nodesMutex.Unlock()
+	for _, node := range nodes {
+		s.nodes[node.NodeId] = node
+	}
```
richardhuaaa, line 163: Can you walk through what happens in the scenario when a listener is notified of a new node, but the registry hasn't written that node into its map yet? (e.g. the listener calls `GetNodes()` first)

dev/up
```diff
@@ -2,15 +2,18 @@
...
 
 go mod tidy
+git submodule update --init --recursive
 
 if ! which forge &>/dev/null; then curl -L https://foundry.paradigm.xyz | bash ; fi
 if ! which migrate &>/dev/null; then brew install golang-migrate; fi
 if ! which golangci-lint &>/dev/null; then brew install golangci-lint; fi
 if ! which shellcheck &>/dev/null; then brew install shellcheck; fi
-if ! which mockery &>/dev/null; then brew install mockery; fi
+if ! which mockery &>/dev/null; then go install github.com/vektra/mockery/v2; fi
```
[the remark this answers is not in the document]
richardhuaaa: Would it be simpler to just specify the version in the install command here, rather than updating the `tools.go` file?

pkg/registry/contractRegistry.go
```diff
@@ -0,0 +1,215 @@
...
+
+	return out, nil
+}
+
+func convertNode(rawNode abis.NodesNodeWithId) Node {
+	// Unmarshal the signing key.
+	// If invalid, mark the config as being invalid as well. Clients should treat the
+	// node as unhealthy in this case
+	signingKey, err := crypto.UnmarshalPubkey(rawNode.Node.SigningKeyPub)
```
richardhuaaa, line 198: I think we should loudly surface an error instead of quietly excluding the node.

It's not expected for an invalid key to be written to the registry, and failing to read it here would imply some inconsistency in how the key is interpreted.

This would also be an argument for why we shouldn't let node operators change the http address themselves. It might be simpler for the DAO to do all validation before writing to the registry (or having the smart contract do the validation) so that everything in the registry can be treated as correct downstream.

pkg/registry/fixedRegistry.go
```diff
@@ -0,0 +1,41 @@
+package registry
+
+import "sync"
+
+// TODO: Delete this or move to a test file
+type FixedNodeRegistry struct {
+	nodes                     []Node
+	newNodeNotifier           *notifier[[]Node]
+	changedNodeNotifiers      map[uint16]*notifier[Node]
+	changedNodeNotifiersMutex sync.Mutex
```
richardhuaaa, line 10: I don't think these notifiers ever get triggered?

richardhuaaa (approved):
This looks really hard to get right, well done

neekolas:
### Merge activity

* **Aug 13, 1:59 PM PDT**: @neekolas started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/xmtp/xmtpd/107).

pkg/registry/fixedRegistry.go
neekolas: This whole thing is getting deleted pretty soon

neekolas:
> Would it be simpler to just specify the version in the install command here, rather than updating the tools.go file?

IDK. Having it all live in the go.mod does make upgrades nicer than having to edit some random script that's easy to be forgotten about. But I don't love either solution.

pkg/registry/contractRegistry.go
neekolas: I can update it to at least log something loudly.

But I do think we need to treat the inputs in the smart contract as somewhat untrusted. Hopefully we'll pre-validate and this will never happen, but that's a human-driven manual process.

pkg/registry/contractRegistry.go
neekolas: I wouldn't expect the listener to need to do that (they already have all the new nodes). But maybe I should move the trigger call to be after just to be safe.

dev/local.env
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+source dev/contracts/.env
+
+export CHAIN_RPC_URL=$DOCKER_RPC_URL  # From contracts/.env
+export NODE_PRIVATE_KEY=$PRIVATE_KEY # From contracts/.env
```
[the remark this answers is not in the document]
neekolas: We use the stuff in `contracts/local.env` to deploy the smart contract locally. Could consolidate, but I'd like to figure out just how complicated all the local smart contract dev stuff is going to be with a full set of contracts and a more sophisticated deployer. My guess is that we're going to have a lot more configuration that is specific to the contracts.

PR #108, opened 2024-08-12
Build xmtpd docker image and push to ghcr

### LINKED ISSUE #104: Publish docker image
We need to set up a Github action that builds `dev/docker/Dockerfile` and publishes it to the Github Container Registry.

You can see an example from one of our other repos [here](https://github.com/xmtp/libxmtp/blob/533efcb221cfa334198a0d0b81131b0175c4f5f2/.github/workflows/deploy-validation-server.yml#L18-L42).

Once we have a Terraform configuration set up which takes the Docker image as a variable, we can use the `xmtp-labs/terraform-deployer@v1` Github action to automatically update the variable and trigger a new Terraform Cloud run on merge to main.

mkysel:
Builds the xmtpd docker image and pushes it to GHCR.

Based on [libxmtp](https://github.com/xmtp/libxmtp/blob/533efcb221cfa334198a0d0b81131b0175c4f5f2/.github/workflows/deploy-validation-server.yml#L18-L42).
The notable difference is that it also builds the image (but does not push) on PRs. Useful for CI.

The resulting image example can already be seen [here](https://github.com/xmtp/xmtpd/actions/runs/10356333475?pr=108)

Fixes #104

### DIFF
--- a/.github/workflows/build-xmtpd.yml
+++ b/.github/workflows/build-xmtpd.yml
@@ -1,0 +1,42 @@
+name: Build xmtpd image
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  push_to_registry:
+    name: Push Docker Image to GitHub Packages
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Log in to the container registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build-xmtpd.yml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          images: ghcr.io/xmtp/xmtpd
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v6
+        id: push
+        with:
+          context: .
+          file: ./dev/docker/Dockerfile
+          push: ${{ github.event_name != 'pull_request' }}
```
neekolas, line 40: Nice touch

neekolas (approved):

PR #112, opened 2024-08-15
Signing and SIDs via Registrant, refactor tests

richardhuaaa:
Apologies, this one exploded in size once I realized I had to refactor the tests. This PR:

1. Creates a 'registrant' struct that holds the signing key and node ID:
   - On creation, verify that the specified private key matches both the registry and what is in the database
   - Issue SID's and sign payloads using the signing key and node ID
2. Refactors the tests:
   - Plumb registrant into existing tests
   - Allow multiple servers to run simultaneously without overwriting each other's state
   - Remove circular dependencies with the new mocks


cmd/replication/main.go
```diff
@@ -40,7 +41,23 @@ func main() {
...
+			log.Fatal("initializing database", zap.Error(err))
+		}
+
+		s, err := server.NewReplicationServer(
+			ctx,
+			log,
+			options,
+			registry.NewFixedNodeRegistry([]registry.Node{}),
+			db,
+		)
```
richardhuaaa, line 60: Pulled the DB out of the replication server for dependency injection purposes. Otherwise, multiple instantiated servers in a test would share the same database and/or have cleanup issues.

pkg/db/queries.sql
```diff
@@ -3,3 +3,11 @@ INSERT INTO staged_originator_envelopes(payer_envelope)
...
 RETURNING
 	*;
+
+-- name: InsertNodeInfo :one
+INSERT INTO node_info(node_id, public_key)
+	VALUES (@node_id, @public_key)
+	RETURNING *;
+
+-- name: SelectNodeInfo :one
+SELECT * FROM node_info WHERE singleton_id = 1;
```
richardhuaaa, line 13: These could have been combined into a single query that fetches the existing record when there is a conflict. Unfortunately, it looks like the methods for doing this in Postgres either impact readability significantly or have weird caveats.

I chose to do this the simple way because server startup performance isn't critical here.

pkg/registrant/registrant.go
```diff
@@ -0,0 +1,125 @@
...
+	record     registry.Node
+	privateKey *ecdsa.PrivateKey
+}
+
+func NewRegistrant(
+	ctx context.Context,
+	db *queries.Queries,
+	nodeRegistry registry.NodeRegistry,
+	privateKeyString string,
+) (*Registrant, error) {
```
richardhuaaa, line 30: Will split this into util methods

pkg/server/server_test.go
```diff
@@ -1,46 +1,56 @@
...
+	require.NoError(t, err)
+
+	registry := mocks.NewMockNodeRegistry(t)
+	registry.On("GetNodes").Return([]r.Node{
+		{NodeID: 1, SigningKey: &privateKey1.PublicKey},
+		{NodeID: 2, SigningKey: &privateKey2.PublicKey},
+	}, nil)
+
+	server1 := NewTestServer(t, dbs[0], registry, privateKey1)
+	server2 := NewTestServer(t, dbs[1], registry, privateKey2)
```
richardhuaaa, line 53: Here we have multiple nodes running with separate DB/private keys but a shared registry

pkg/registry/contractRegistry_test.go
```diff
@@ -1,4 +1,4 @@
-package registry
+package registry_test
```
richardhuaaa, line 1: Needed to remove circular dependencies - otherwise this test (in the registry package) depends on the mocks package, which depends on the registry package

pkg/db/queries.sql
neekolas: Why does this need to live in the database at all? Can't the `NodeRegistry` just serve as the source of truth here?

cmd/replication/main.go
neekolas: That makes sense

pkg/utils/sid.go
```diff
@@ -1,18 +1,28 @@
 package utils
 
+// Operations to marshall to/from SIDs.
+// It is recommended not to use these directly - instead use a wrapper type that
+// performs the relevant error handling.
+
 const (
-	nodeIDMask  uint64 = 0xFFFF << 48
+	nodeIDBits  uint64 = 16
```
neekolas, line 8: It would be good to have some really clear comments around how these IDs are supposed to work, since bitmasks are not the most readable.

neekolas (approved):

pkg/db/queries.sql
richardhuaaa: Yes, purely so that you get an error if you change either piece of information with the same DB. I imagine it's an easy mistake for partners to make

neekolas: OK. Good to be defensive here. Could definitely see it being an easy mistake to make.

PR #113, opened 2024-08-21
Add more queries, fix DB sequence ID's

richardhuaaa:
I've fixed a number of issues, would like to suggest some best practices going forward that I've implemented here:

1. **Reliable sequence ID's**. Whenever we add a table that uses a sequence for ordering, we should also add a Postgres stored function immediately below it that performs insertions by taking an advisory lock first. This ensures that inserts on the table are [processed serially](https://espadrine.github.io/blog/posts/two-postgresql-sequence-misconceptions.html#Order_violation).
2. **Idempotent inserts**. Whenever we add an insert query, we should add an `ON CONFLICT DO NOTHING` clause, and either return the inserted count via `:execrows` or the row itself via `:one`. This saves the caller from having to parse untyped errors in search of unique constraint violations, which are sometimes expected.
3. **Result granularity**.
    - Postgres functions should always return the inserted rows via `RETURNING *`, even if the result is not currently used. This is because any future changes to the function would require a DB migration. 
    - In `queries.sql`, results should be defined as narrowly as possible - for example, do not return the row via `:one` if the code is not using it. This saves networking bandwidth, and the query can easily be changed later on without a migration.
4. **Consistent SID terminology**. Settling on the terms `Node ID` and `Sequence ID` as the components of a `SID`, removing references to `LocalID`.

Some other stuff I've done:

1. I've split up the `sid` field in the `gateway_envelopes` table into `node_id` and `sequence_id`, because the postgres `BIGINT` type is a signed 64-bit int that cannot hold a `uint64` without bad side effects.
2. I've added the `insert_gateway_envelope` and `delete_staged_originator_envelope` queries which are needed for the stacked PR's.

### DIFF
--- a/pkg/migrations/00001_init-schema.up.sql
+++ b/pkg/migrations/00001_init-schema.up.sql
@@ -3,7 +3,6 @@
 CREATE TABLE node_info(
 	node_id INTEGER NOT NULL,
 	public_key BYTEA NOT NULL,
-
 	singleton_id SMALLINT PRIMARY KEY DEFAULT 1,
 	CONSTRAINT is_singleton CHECK (singleton_id = 1)
 );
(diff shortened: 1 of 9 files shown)


richardhuaaa:
* **#115** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/115?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#114** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/114?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#113** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/113?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pkg/registrant/registrant.go
```diff
@@ -121,14 +121,18 @@ func getRegistryRecord(
...
 		queries.InsertNodeInfoParams{
 			NodeID:    int32(record.NodeID),
 			PublicKey: crypto.FromECDSAPub(record.SigningKey),
 		},
 	)
 	if err != nil {
+		return fmt.Errorf("unable to insert node info into database: %v", err)
+	}
+
+	if numRows == 0 {
```
richardhuaaa, line 135: An example of where `ON CONFLICT DO NOTHING` + `:execRows` is useful

pkg/migrations/00001_init-schema.up.sql
```diff
@@ -12,15 +11,35 @@ CREATE TABLE node_info(
...
 -- Node queries
-CREATE UNIQUE INDEX idx_gateway_envelopes_originator_sid ON gateway_envelopes(originator_sid);
+CREATE UNIQUE INDEX idx_gateway_envelopes_originator_sid ON gateway_envelopes(originator_id, sequence_id);
 
+CREATE FUNCTION insert_gateway_envelope(originator_id INT, sequence_id BIGINT, topic BYTEA, originator_envelope BYTEA)
+	RETURNS SETOF gateway_envelopes
+	AS $$
+BEGIN
+	-- Ensures that the generated sequence ID matches the insertion order
+	-- Only released at the end of the enclosing transaction - beware if called within a long transaction
```
richardhuaaa, line 31: Every SQL statement gets wrapped into a transaction by default, but what we need to be careful of is if our application code creates an 'enclosing transaction' that performs a bunch of stuff after this function is called.

If that is needed, the application code can manually wrap the function call in a sub-transaction, to make sure the lock is released in a timely manner.

neekolas:
Great work going deep on these sequencing problems. Unfortunate that we have to use table-level locks, but I don't see much of an alternative.

This [article](https://mattjames.dev/auto-increment-ids-are-not-strictly-monotonic/) is also useful and lists a bunch of alternative solutions to the problem (and why they're all bad). Batching would help a bit, but only if we had enough volume to have large batches in very short time windows. We could always add batching later.

neekolas (approved):

PR #114, opened 2024-08-21
Add DB subscription

richardhuaaa:
Adds a struct that can subscribe to changes in a database query, starting from some cursor onwards. Features:
- Poll based on an interval or on an input notifier channel
- Serve updates via an output channel
- Stops polling when context is closed
- Non-racy unit tests (hopefully)

Also adds a flexible DB query against the gateway_envelopes table that can query based on a range of params. Given the size of the query, I should add a unit test for it eventually.


richardhuaaa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/114?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#115** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/115?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#114** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/114?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#113** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/113?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

pkg/db/subscription.go
```diff
@@ -0,0 +1,92 @@
...
+			panic("TODO(rich)")
+		}
+
+		for {
+			select {
+			case <-s.options.Notifier:
+				if err = s.poll(); err != nil {
+					panic("TODO(rich)")
+				}
+			case <-time.After(s.options.Interval):
```
neekolas, line 56: `time.After` is a Go feature that's a bit of a trap. `time.Ticker` is [weirdly more performant](https://github.com/ChainSafe/gossamer/issues/1623) and doesn't have [known memory leaks](https://100go.co/#timeafter-and-memory-leaks-76).

pkg/db/queries.sql
```diff
@@ -14,7 +14,22 @@ WHERE
...
+-- name: SelectGatewayEnvelopes :many
+SELECT
+	*
+FROM
+	gateway_envelopes
+WHERE (sqlc.narg('topic')::BYTEA IS NULL
+	OR topic = @topic)
+AND (sqlc.narg('originator_node_id')::INT IS NULL
+	OR originator_node_id = @originator_node_id)
+AND (sqlc.narg('originator_sequence_id')::BIGINT IS NULL
```
mkysel, line 28: unrelated to the PR, but who generates the sequence ID?

richardhuaaa: Great question - the originator does, this would be extracted from the `originator_sid` field of `UnsignedOriginatorEnvelope`.

As for how the originator allocates these ID's in the first place, it is created by the `BIGSERIAL id` field of the `staged_originator_envelope` table - so we're using a postgres sequence for this.

pkg/db/subscription.go
```diff
@@ -0,0 +1,101 @@
...
+
+		timer := time.NewTimer(s.options.Interval)
+		for {
+			timer.Reset(s.options.Interval)
+			select {
+			case <-s.ctx.Done():
+				s.log.Info("Context done; stopping subscription")
+				close(s.updates)
+				return
+			case <-s.options.Notifier:
```
neekolas, line 67: I think you can reset the timer here, so that if the timer is a 2s interval and you get notified after 1.8s, it won't check again in 0.2s

pkg/db/subscription.go
```diff
@@ -0,0 +1,101 @@
...
+				s.poll()
+			case <-timer.C:
+				s.poll()
```
neekolas, line 70: You'll need to reset the timer here, or it'll only fire one time

neekolas (approved):

pkg/db/subscription.go
richardhuaaa: See other comment!

pkg/db/subscription.go
richardhuaaa: The timer is reset at the start of the `for` loop, regardless of which branch it went down previously - I think that addresses what you're asking about, but let me know if I misunderstood

neekolas: Totally missed that line

PR #115, opened 2024-08-21
Add publish worker

richardhuaaa:
- Adds a publish worker that performs in-order insertion into the `gateway_envelopes` table
- Adds basic validation of the client envelope on publish
- Store the topic on the `staged_originated_envelopes` table - we extract this during the API call, so that the publish worker doesn't need to do any additional unmarshaling or validation.

Would particularly love feedback on the error handling in the worker, and if there's any test cases I should add to `service_test.go`.

### DIFF
--- a/pkg/api/publishWorker.go
+++ b/pkg/api/publishWorker.go
@@ -1,0 +1,147 @@
+package api
+
+import (
+	"context"
+	"database/sql"
+	"time"
+
+	"github.com/xmtp/xmtpd/pkg/db"
+	"github.com/xmtp/xmtpd/pkg/db/queries"
+	"github.com/xmtp/xmtpd/pkg/registrant"
+	"go.uber.org/zap"
+	"google.golang.org/protobuf/proto"
+)
+
+type PublishWorker struct {
+	ctx          context.Context
+	log          *zap.Logger
+	listener     <-chan []queries.StagedOriginatorEnvelope
+	notifier     chan<- bool
+	registrant   *registrant.Registrant
+	store        *sql.DB
+	subscription db.DBSubscription[queries.StagedOriginatorEnvelope]
+}
+
+func StartPublishWorker(
+	ctx context.Context,
+	log *zap.Logger,
+	reg *registrant.Registrant,
+	store *sql.DB,
+) (*PublishWorker, error) {
+	q := queries.New(store)
+	query := func(ctx context.Context, lastSeenID int64, numRows int32) ([]queries.StagedOriginatorEnvelope, int64, error) {
+		results, err := q.SelectStagedOriginatorEnvelopes(
+			ctx,
+			queries.SelectStagedOriginatorEnvelopesParams{
+				LastSeenID: lastSeenID,
+				NumRows:    numRows,
+			},
+		)
+		if err != nil {
+			return nil, 0, err
+		}
+		if len(results) > 0 {
+			lastSeenID = results[len(results)-1].ID
+		}
+		return results, lastSeenID, nil
+	}
+	notifier := make(chan bool, 1)
+	subscription := db.NewDBSubscription(
+		ctx,
+		log,
+		query,
+		0, // lastSeenID
+		db.PollingOptions{Interval: time.Second, Notifier: notifier, NumRows: 100},
+	)
+	listener, err := subscription.Start()
+	if err != nil {
+		return nil, err
+	}
+
+	worker := &PublishWorker{
+		ctx:          ctx,
+		log:          log,
+		notifier:     notifier,
+		subscription: *subscription,
+		listener:     listener,
+		registrant:   reg,
+		store:        store,
+	}
+	go worker.start()
+
+	return worker, nil
+}
+
+func (p *PublishWorker) NotifyStagedPublish() {
+	select {
+	case p.notifier <- true:
+	default:
+	}
+}
+
+func (p *PublishWorker) start() {
+	for {
+		select {
+		case <-p.ctx.Done():
+			return
+		case new_batch := <-p.listener:
+			for _, stagedEnv := range new_batch {
+				for !p.publishStagedEnvelope(stagedEnv) {
+					// Infinite retry on failure to publish; we cannot
+					// continue to the next envelope until this one is processed
+					time.Sleep(time.Second)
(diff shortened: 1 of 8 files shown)


richardhuaaa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/115?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#115** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/115?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#114** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/114?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#113** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/113?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

pkg/api/publishWorker.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func StartPublishWorker(
+	ctx context.Context,
+	log *zap.Logger,
+	reg *registrant.Registrant,
+	store *sql.DB,
+) (*PublishWorker, error) {
+	query := func(ctx context.Context, lastSeenID int64, numRows int32) ([]queries.StagedOriginatorEnvelope, int64, error) {
+		results, err := queries.New(store).SelectStagedOriginatorEnvelopes(
```
neekolas, line 32: Why do we need to re-instantiate the `Queries` struct on each call? Can't we just instantiate it outside of the callback?

pkg/api/publishWorker.go
```diff
@@ -0,0 +1,147 @@
...
+	case p.notifier <- true:
+	default:
+	}
+}
+
+func (p *PublishWorker) start() {
+	for {
+		select {
+		case <-p.ctx.Done():
+			close(p.notifier)
```
neekolas, line 85: Sending on a closed channel will cause a panic, so we want to make sure that the sender is in charge of closing the channel.

In this case, an API request might still come in after the context is canceled that calls `worker.NotifyStagedPublish`, which would cause a panic.

I'd suggest we make the notifier channel get passed in to the worker from the ApiService - and have the ApiService responsible for writing to the channel - so that the API service can be responsible for cleaning up in the right order during shutdown.

pkg/api/publishWorker.go
```diff
@@ -0,0 +1,147 @@
...
+		)
+		return false
+	}
+	originatorBytes, err := proto.Marshal(originatorEnv)
+	if err != nil {
+		logger.Error("Failed to marshal originator envelope", zap.Error(err))
+		return false
+	}
+
+	q := queries.New(p.store)
```
neekolas, line 115: Do we actually want to be wrapping this in a transaction, so that the insert can never happen without the corresponding delete?

richardhuaaa: That was actually my initial plan, but I realized that this could work without the transaction. Basically, the worst case scenario if a delete doesn't go through is that another server tries to do the same insert (and gets a unique constraint error).

The tradeoff here is potentially wasting server resources in edge cases, but in exchange you get better average throughput and simplicity due to not needing transactions. Would love a second pair of eyes on this reasoning though.

pkg/api/publishWorker.go
```diff
@@ -0,0 +1,147 @@
...
+		queries.InsertGatewayEnvelopeParams{
+			OriginatorID:         int32(p.registrant.NodeID()),
+			OriginatorSequenceID: stagedEnv.ID,
+			Topic:                stagedEnv.Topic,
+			OriginatorEnvelope:   originatorBytes,
+		},
+	)
+	if err != nil {
+		logger.Error("Failed to insert gateway envelope", zap.Error(err))
+		return false
```
richardhuaaa, line 129: There is a risk here that we infinitely retry on this step or one of the steps above in an unanticipated error case. I think that we should be okay here but a second pair of eyes can't hurt

pkg/api/publishWorker.go
neekolas: That does sound like it checks out. It's going to keep showing up in the `SelectStagedOriginatorEnvelopes` query until it gets deleted

pkg/api/publishWorker.go
neekolas: Well, i guess it won't panic because `NotifyStagedPublish` sends in a select statement.

Still feels like we should try and close it from the sender side. That's almost always the right pattern in Go.

pkg/api/publishWorker.go
neekolas: I don't know why the proto marshaling would fail, but if it did we would be in an endless loop of retries here.

Would probably have to be a maliciously crafted payload

pkg/api/publishWorker.go
neekolas: If the channel is only getting closed when the server shuts down, maybe we don't need to close it at all.

Because you can have multiple concurrent senders it's going to be hard for the `ApiService` to know when it's fully safe to close. Given that only happens when the server is shutting down anyways, it should be fine.

richardhuaaa: Good point, just not closing the channel at all. The nice thing about `NotifyStagedPublish` is that the notifier channel is self-contained in the `PublishWorker`, so an alternative would have been to check/close that channel inside that method instead

pkg/api/publishWorker.go
richardhuaaa: From my research, it seems like it fails if some expected fields of the proto are not set. In this case, the proto is `OriginatorEnvelope` with two fields that we are setting [here](https://github.com/xmtp/xmtpd/blob/a8cef43a30f683da94f79e36cb71f81225fe85bd/pkg/registrant/registrant.go#L93), so I expect that we should be able to prevent this error happening.

I still think we should catch the error just in case though, firstly so that we can detect it, secondly so that we don't screw up payload ordering if there is an unanticipated error type that is super common.

neekolas (approved):

PR #116, opened 2024-08-26
Setup golines checker

mkysel:
Make sure that CI catches files that have not been formatted by `golines`.
Add a script that does the formatting and ignores autogenerated files.

Both idea and vscode will reformat a generated file if you save it. Since golines does not offer a good way to ignore these files correctly.

Here is the CI detecting a poorly formatted file: https://github.com/xmtp/xmtpd/actions/runs/10564716546/job/29267709565


pkg/config/options.go
```diff
@@ -9,29 +9,29 @@ type ApiOptions struct {
 }
 
 type ContractsOptions struct {
```
richardhuaaa, line 11: Could optionally exclude this file/directory too, up to you

richardhuaaa (approved):
Thanks for tackling this!

PR #117, opened 2024-08-26
Unify port

mkysel:
Pre-PR we expose a different port than the docker image expects.

I picked `80` since it makes the most sense, but we can pick any other port if people have opinions.

### DIFF
--- a/pkg/config/options.go
+++ b/pkg/config/options.go
@@ -5,7 +5,7 @@ import (
 )
 
 type ApiOptions struct {
-	Port int `short:"p" long:"port" description:"Port to listen on" default:"5050"`
+	Port int `short:"p" long:"port" description:"Port to listen on" default:"80"`
 }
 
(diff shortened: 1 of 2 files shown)


pkg/config/options.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 type ApiOptions struct {
-	Port int `short:"p" long:"port" description:"Port to listen on" default:"5050"`
+	Port int `short:"p" long:"port" description:"Port to listen on" default:"80"`
```
neekolas, line 8: I know we talked about 80 earlier, but for defaults that get used for local dev we should really use a unique port. If every XMTP service defaulted to 80, we'd constantly have port conflicts doing anything locally.

Totally OK with setting it to 80 in the production service, which we can do with the `--port` command line argument at runtime.

neekolas (approved):

PR #120, opened 2024-08-29
Save blockchain message

neekolas:
## tl;dr

- Implements the LogStorer interface to save messages from the blockchain
- Adds test helpers to get dynamic config without relying on environment variables


neekolas:
* **#120** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/120?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pkg/indexer/storer/groupMessage.go
```diff
@@ -2,23 +2,55 @@ package storer
...
-	"errors"
+	"fmt"
 
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/xmtp/xmtpd/pkg/abis"
 	"github.com/xmtp/xmtpd/pkg/db/queries"
 	"go.uber.org/zap"
 )
 
 type GroupMessageStorer struct {
```
richardhuaaa, line 13: Understanding the `LogStorer` interface a little better - what are some other examples of `LogStorer` we might want, now or in the future?

pkg/indexer/storer/groupMessage.go
```diff
@@ -2,23 +2,55 @@ package storer
...
+		return NewLogStorageError(err, false)
+	}
+
+	// TODO:nm figure out topic structure
+	topic := buildTopic(msgSent.GroupId)
+
+	s.logger.Debug("Inserting message from contract", zap.String("topic", topic))
+
+	if _, err = s.queries.InsertGatewayEnvelope(ctx, queries.InsertGatewayEnvelopeParams{
+		// We may not want to hardcode this to 0 and have an originator ID for each smart contract?
```
richardhuaaa, line 40: Understanding this a little bit better - what other smart contracts are we envisioning here? Is this related to sharding?

pkg/indexer/storer/groupMessage.go
```diff
@@ -2,23 +2,55 @@ package storer
...
+		OriginatorID:         0,
+		OriginatorSequenceID: int64(msgSent.SequenceId),
+		Topic:                []byte(topic),
+		OriginatorEnvelope:   msgSent.Message, // TODO:nm parse originator envelope and do some validation
```
richardhuaaa, line 44: We can probably share this validation code, if you don't want to deal with it

pkg/indexer/storer/groupMessage.go
```diff
@@ -2,23 +2,55 @@ package storer
...
+	if err != nil {
+		return NewLogStorageError(err, false)
+	}
+
+	// TODO:nm figure out topic structure
+	topic := buildTopic(msgSent.GroupId)
+
+	s.logger.Debug("Inserting message from contract", zap.String("topic", topic))
+
+	if _, err = s.queries.InsertGatewayEnvelope(ctx, queries.InsertGatewayEnvelopeParams{
```
richardhuaaa, line 39: Is it right that we don't ever expect to process the same log twice? If so, is it problematic if we run into a unique constraint error here (which would mean the (originator ID, sequence ID) tuple already exists)?

richardhuaaa (approved):

pkg/indexer/storer/groupMessage.go
neekolas: Probably just this and `IdentityUpdate`s in the near future.

pkg/indexer/storer/groupMessage.go
neekolas: We could if we have two instances of the replication node running, and both are indexing the chain.

If a node gets shut down mid-block, it'll also restart at the top of that block.

The only hard constraint is that they are both processing messages in ascending order.

pkg/indexer/storer/groupMessage.go
neekolas: And maybe upgradeability. At some point we are going to replace the smart contract and we have some more optionality if we've designed around it having its own originator ID where we can handle payloads differently.

Smart contract cutover would be a complicated migration to synchronize and I don't have all the answers on how to do it. But more options feels better.

neekolas:
### Merge activity

* **Aug 29, 4:23 PM PDT**: @neekolas started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/xmtp/xmtpd/120).

PR #128, opened 2024-08-30
Allow setting options via ENV

mkysel:
Add the option to set any option via the ENV.

Prefix all variables with `XMTPD` as best practices dictate.

Remove manual env parsing.


mkysel:
I will undo the mockery changes...

neekolas (approved):
Thanks for tackling this. Much tidier this way

neekolas:
> I will undo the mockery changes...

If you run `dev/update-tools` it should put you on the same version of mockery as the go.mod. You may have to `brew uninstall` as well, depending on the order of your PATH

mkysel:
> > I will undo the mockery changes...
> 
> If you run `dev/update-tools` it should put you on the same version of mockery as the go.mod. You may have to `brew uninstall` as well, depending on the order of your PATH

Thanks!

PR #129, opened 2024-08-30
Add metrics and tracing

mkysel:
Most of this is copied from `node-go` with some slight adjustments.

Expose a metrics server on a port even though no custom metrics are provided yet.


mkysel:
so far this is untested. I am having some issues getting the server started. Won't merge anything until then.

dev/run
```diff
@@ -4,4 +4,4 @@ set -eu
 
 . dev/local.env
 
-go run cmd/replication/main.go
\ No newline at end of file
+go run cmd/replication/main.go "$@"
```
mkysel, line 7: allows passing of args for easy testing such as:
```
./dev/run --metrics.enable --tracing.enable
```

cmd/replication/main.go
```diff
@@ -49,17 +61,34 @@ func main() {
...
-			log.Fatal("initializing database", zap.Error(err))
+			logger.Fatal("initializing database", zap.Error(err))
 		}
 
 		privateKey, err := utils.ParseEcdsaPrivateKey(options.SignerPrivateKey)
 		if err != nil {
 			log.Fatal("parsing private key", zap.Error(err))
 		}
 
+		var mtcs *metrics.Server
```
neekolas, line 72: I don't have strong opinions on this, but it would be fine to create the metrics server inside `pkg/server`. It might make more logical sense given that the server is responsible for shutting down the metrics.

neekolas (approved):

PR #131, opened 2024-08-30
Implement query endpoint

### LINKED ISSUE #124: Query endpoint using existing DB query

richardhuaaa:
This implements the query endpoint and tests. I haven't implemented support for queries from a vector clock cursor yet due to the changes in https://github.com/xmtp/proto/pull/201/files, but will implement support for it in the next PR.

Closes #124


richardhuaaa:
* **#131** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/131?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pkg/db/subscription.go
```diff
@@ -79,7 +79,9 @@ func (s *DBSubscription[ValueType]) poll() {
 	// Repeatedly query page by page until no more results
 	for {
 		results, lastID, err := s.query(s.ctx, s.lastSeenID, s.options.NumRows)
-		if err != nil {
+		if s.ctx.Err() != nil {
+			break
+		} else if err != nil {
```
richardhuaaa, line 84: It's pretty benign, but whenever I look at logs for the tests, they always end with entries saying `Error querying for DB subscription`. I assume this is happening during shutdown. This change was meant to silence those logs, but it didn't silence them - I suspect we're shutting things down in the wrong order. I can clean up the logs at a future date

neekolas: I do have a PR waiting in the wings that should simplify our shutdown flow. Not sure if it fixes this particular issue, but we can check after.

richardhuaaa:
I've incorporated the changes from https://github.com/xmtp/proto/pull/201/files into this PR, hence the large line count, which is mostly generated code.

I've left queries using a vector clock cursor unimplemented in this PR, as it's more involved - will implement that as a separate PR.

pkg/api/publishWorker.go
```diff
@@ -134,7 +136,9 @@ func (p *PublishWorker) publishStagedEnvelope(stagedEnv queries.StagedOriginator
 
 	// Try to delete the row regardless of if the gateway envelope was inserted elsewhere
 	deleted, err := q.DeleteStagedOriginatorEnvelope(context.Background(), stagedEnv.ID)
```
neekolas, line 138: We can also pass in `p.ctx` here so that this will fail if the server is in the process of shutting down

richardhuaaa: Totally missed this one, good catch

neekolas (approved):

PR #134, opened 2024-09-03
Tag by SHA

mkysel:
Automatically tag a docker image by it's git sha.

This is useful for pinning and debugging.

For example:
```
Docker tags
  ghcr.io/xmtp/xmtpd:pr-134
  ghcr.io/xmtp/xmtpd:sha-56e2af4
```

Drive-by allow passing args to `dev/run` such as `./dev/run --api.port=1000`

### DIFF
--- a/.github/workflows/build-xmtpd.yml
+++ b/.github/workflows/build-xmtpd.yml
@@ -14,12 +14,11 @@ jobs:
     permissions:
       contents: read
       packages: write
+    env:
+      DOCKER_METADATA_PR_HEAD_SHA: true
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-        # we want the HEAD commit instead of the merge SHA for the docker tag
-        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-xmtpd.yml
```diff
@@ -14,6 +14,8 @@ jobs:
     permissions:
       contents: read
       packages: write
+    env:
+      DOCKER_METADATA_PR_HEAD_SHA: true
```
richardhuaaa, line 18: Noob question - the [doc](https://github.com/docker/metadata-action?tab=readme-ov-file#environment-variables) for this says:
```
If true, set associated head SHA instead of commit SHA that triggered the workflow on pull request event
```
What is the distinction here between the head SHA and the commit SHA?

richardhuaaa (approved):
Nice improvements!

PR #145, opened 2024-09-05
Rework schema and queries for gateway envelopes

richardhuaaa:
Now that we have [vector clock cursors](https://github.com/xmtp/proto/pull/201), removing the need for a gateway sequence ID, there are some schema changes that are worth going over with a keen eye:
- Remove the `sequence_id` on `gateway_envelopes` and replace it with a timestamp
   - The timestamp needs to match the insertion order *per-originator*, but does not need to match insertion order globally.
- Updating the `SelectGatewayEnvelopes` query to be vector clock aware (only returning envelopes greater than the vector clock cursor)

Because the any changes in this PR affect the downstream work, I'm splitting the work to make the Go code vector-clock aware into a separate PR. Instead, I've left TODO's in the code that I will replace in the future PR.

#132


pkg/db/queries.sql
```diff
@@ -13,23 +13,36 @@ WHERE
...
+		UNNEST(@cursor_node_ids::INT[]) AS cursor_node_id,
+		UNNEST(@cursor_sequence_ids::BIGINT[]) AS cursor_sequence_id
+)
 SELECT
-	*
+	gateway_envelopes.*
 FROM
 	gateway_envelopes
+	-- Assumption: There is only one cursor per node ID. Caller must verify this
+	LEFT JOIN cursors ON gateway_envelopes.originator_node_id = cursors.cursor_node_id
```
richardhuaaa, line 32: The most important changes are in this query. We're doing this interesting JOIN operation against the cursor specified by the caller.

I originally wanted to use a different approach of specifying one WHERE clause per cursor entry, but as far as I can tell [sqlc does not support this](https://github.com/sqlc-dev/sqlc/issues/3578).

The JOIN method is more readable IMO, although I'm unsure of the performance implication - it depends on how much optimization postgres does internally. When we are further along in the project, we can do some benchmarking with `explain analyze` etc.

pkg/db/queries.sql
```diff
@@ -13,23 +13,36 @@ WHERE
...
-AND (sqlc.narg('originator_sequence_id')::BIGINT IS NULL
-	OR originator_sequence_id > @originator_sequence_id)
-AND (sqlc.narg('gateway_sequence_id')::BIGINT IS NULL
-	OR id > @gateway_sequence_id)
+AND (cursor_sequence_id IS NULL
+	OR originator_sequence_id > cursor_sequence_id)
+ORDER BY
+	-- Assumption: envelopes are inserted in sequence_id order per originator, therefore
+	-- gateway_time preserves sequence_id order
+	gateway_time,
```
richardhuaaa, line 43: This sorts the envelopes roughly in the order they were received. The reason this should match sequence_id order is that:
- For envelopes originated elsewhere, we are inserting them sequentially from the stream
- For envelopes originated by ourselves, we are inserting them sequentially via the publish worker

There are some alternative sort methods I considered:
- Sort by originator time. The advantage is the ordering is the same on all gateways.
- Sort by originator node ID/originator sequence ID.

Sorting by gateway_time has the advantage that clients receive payloads in the same order from the gateway regardless of the time that they chose to sync (you cannot receive a payload later on that would have been inserted earlier in the sort order). Note that if we want to sort by originator timestamp in the case of causal ordering tiebreaks, clients can still do that locally. I can definitely be convinced otherwise on this decision though.

richardhuaaa:
I also considered whether or not we could remove the `staged_originator_envelopes` table and publish worker, and simply add a postgres function that acquires an advisory lock before assigning the `originator_sequence_id` when inserting into `gateway_envelopes`. I decided against, because the sequence ID needs to be signed, and holding an advisory lock for the duration of a signing operation does not seem scalable.

pkg/db/queries.sql
neekolas: If it turns out to be a big deal, there is always the option of just constructing the parameterized SQL statement outside of sqlc and running it against the DB driver.

pkg/db/queries.sql
```diff
@@ -13,23 +13,36 @@ WHERE
...
+	originator_node_id,
+	originator_sequence_id ASC
```
neekolas, line 45: Do we actually need these secondary sorts. Cheaper to index if it's just one field, and it doesn't seem to matter given that ties will be rare.

neekolas (approved):

pkg/db/queries.sql
richardhuaaa: I worry that consecutive payloads from the same originator could have the same timestamp. Maybe when we analyze the query performance later we can remove it if it's costly? Should be an easy change

neekolas: Is there an advantage to doing it this way vs using the table's `tid` as a tiebreaker (the Postgres default in ties)?

https://www.postgresql.org/docs/current/indexes-ordering.html

richardhuaaa: I'm not sure to be honest. This is the definition of `tid` I found:

```
A final identifier type used by the system is tid, or tuple identifier (row identifier). This is the data type of the system column ctid. A tuple ID is a pair (block number, tuple index within block) that identifies the physical location of the row within its table.
```

It seems somewhat tied to the implementation. I'm inclined to write the query first in the naive/intuitive way, and determine later on whether or not there is a performance impact - I'm somewhat hopeful that the query planner will optimize it seeing as we won't have ties most of the time.

richardhuaaa: Open to feedback though! Definitely not my expertise.

neekolas: Query planner is pretty smart. You might be right it'll just use the index on the first column in the ORDER BY and reshuffle in memory for ties.

We can find out when we get to optimizations and profiling

PR #153, opened 2024-09-10
Make server logic vector clock-aware

richardhuaaa:
- Modify server logic for queries and subscriptions to support vector clock cursors
- Implement the changes from https://github.com/xmtp/proto/pull/206, including splitting SID into two fields, using a dict for the vector clock, and modifying node ID from uint16 to uint32

I was worried about type-conversions between int32 and uint32 and overflow issues, but it looks like with two's complement, [overflowed int32's behave the same as uint32's after conversion](https://go.dev/play/p/pQmLPP56nzx) (or in other words, the increment operation does exactly the same thing to the underlying byte representation either way). In other words, our postgres schema can only support signed int32's, but it seems like we can just pretend they are uint32's and have the same effect.

We do still need to worry about the postgres sequence overflowing, as well as the max value of uint32/uint64 - when implementing the node-to-node syncing logic, my plan is to regularly log the node's vector clock, so that we can see it coming from a long way away.


richardhuaaa:
* **#153** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/153?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pkg/db/subscription.go
```diff
@@ -8,7 +8,11 @@ import (
 	"go.uber.org/zap"
 )
 
-type PollableDBQuery[ValueType any] func(ctx context.Context, lastSeenID int64, numRows int32) (results []ValueType, lastID int64, err error)
+type PollableDBQuery[ValueType any, CursorType any] func(
+	ctx context.Context,
+	lastSeen CursorType,
+	numRows int32,
+) (results []ValueType, nextCursor CursorType, err error)
```
richardhuaaa, line 15: Main change is here - `PollableDBQuery` now uses a generic cursor type rather than an `int64`

pkg/api/service.go
```diff
@@ -16,7 +16,8 @@ import (
 )
 
 const (
-	maxRequestedRows int32 = 1000
+	maxRequestedRows     uint32 = 1000
+	maxVectorClockLength int    = 100
```
richardhuaaa, line 20: I added this mainly to stop malicious actors from flooding our DB with large requests using huge vector clock queries. We could theoretically make this bigger

neekolas (approved):

pkg/api/service.go
neekolas: In practice, this should always be < the number of nodes in the registry. Maybe that can be the upper bound?

PR #160, opened 2024-09-12
Non-thread-safe subscriptions

richardhuaaa:
Introduces a basic initial version of client subscriptions. A worker maintains a global database subscription, and pushes updates to relevant client listeners.

There are many pieces coming in later PR's:

- Thread safety for adding and reading listeners
- Listener cleanup, esp when channels are full
- Subscribing from a vector clock cursor (which also gives a way to recover from backpressure)
- Logging and load/performance metrics
- Bidirectional streaming

#125

### DIFF
--- a/pkg/api/service.go
+++ b/pkg/api/service.go
@@ -85,6 +85,10 @@ func (s *Service) BatchSubscribeEnvelopes(
 	}
 
 	ch, err := s.subscribeWorker.subscribe(requests)
+	if err != nil {
+		// TODO(rich) Tidy error interface, validate before sending header
+		return err
+	}
 	defer func() {
 		// TODO(rich) Handle unsubscribe
 		// if sub != nil {
--- a/pkg/api/service.go
+++ b/pkg/api/service.go
@@ -66,9 +66,7 @@ func (s *Service) BatchSubscribeEnvelopes(
 	req *message_api.BatchSubscribeEnvelopesRequest,
 	stream message_api.ReplicationApi_BatchSubscribeEnvelopesServer,
 ) error {
-	log := s.log.Named("subscribe") // .With(zap.Strings("content_topics", req.ContentTopics))
-	log.Debug("started")
-	defer log.Debug("stopped")
+	log := s.log.Named("subscribe")
 
 	// Send a header (any header) to fix an issue with Tonic based GRPC clients.
 	// See: https://github.com/xmtp/libxmtp/pull/58
@@ -122,6 +120,7 @@ func (s *Service) QueryEnvelopes(
 	ctx context.Context,
 	req *message_api.QueryEnvelopesRequest,
 ) (*message_api.QueryEnvelopesResponse, error) {
+	log := s.log.Named("query")
 	params, err := s.queryReqToDBParams(req)
 	if err != nil {
 		return nil, err
@@ -138,7 +137,7 @@ func (s *Service) QueryEnvelopes(
 		err := proto.Unmarshal(row.OriginatorEnvelope, originatorEnv)
 		if err != nil {
 			// We expect to have already validated the envelope when it was inserted
-			s.log.Error("could not unmarshal originator envelope", zap.Error(err))
+			log.Error("could not unmarshal originator envelope", zap.Error(err))
 			continue
 		}
 		envs = append(envs, originatorEnv)
@@ -152,7 +151,6 @@ func (s *Service) QueryEnvelopes(
 func (s *Service) queryReqToDBParams(
 	req *message_api.QueryEnvelopesRequest,
 ) (*queries.SelectGatewayEnvelopesParams, error) {
-	// TODO(rich) named logs
 	params := queries.SelectGatewayEnvelopesParams{
 		Topic:             nil,
 		OriginatorNodeID:  sql.NullInt32{},
--- a/pkg/api/service.go
+++ b/pkg/api/service.go
@@ -66,9 +66,6 @@ func (s *Service) BatchSubscribeEnvelopes(
 	req *message_api.BatchSubscribeEnvelopesRequest,
 	stream message_api.ReplicationApi_BatchSubscribeEnvelopesServer,
 ) error {
-	// TODO(rich): Figure out subscribe2
-	// TODO(rich): Allow subscription to be updated
-
 	log := s.log.Named("subscribe") // .With(zap.Strings("content_topics", req.ContentTopics))
 	log.Debug("started")
 	defer log.Debug("stopped")
@@ -85,7 +82,7 @@ func (s *Service) BatchSubscribeEnvelopes(
 		return status.Errorf(codes.InvalidArgument, "missing requests")
 	}
 
-	ch, err := s.subscribeWorker.subscribe(requests)
+	ch, err := s.subscribeWorker.addListeners(requests)
 	if err != nil {
 		// TODO(rich) Tidy error interface, validate before sending header
 		return err
--- a/pkg/api/service.go
+++ b/pkg/api/service.go
@@ -3,12 +3,14 @@ package api
 import (
 	"context"
 	"database/sql"
+	"sync"
 
 	"github.com/xmtp/xmtpd/pkg/db"
 	"github.com/xmtp/xmtpd/pkg/db/queries"
 	"github.com/xmtp/xmtpd/pkg/proto/xmtpv4/message_api"
 	"github.com/xmtp/xmtpd/pkg/registrant"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 
@@ -23,11 +25,12 @@ const (
 type Service struct {
 	message_api.UnimplementedReplicationApiServer
 
-	ctx        context.Context
-	log        *zap.Logger
-	registrant *registrant.Registrant
-	store      *sql.DB
-	worker     *PublishWorker
+	ctx             context.Context
+	log             *zap.Logger
+	registrant      *registrant.Registrant
+	store           *sql.DB
+	publishWorker   *publishWorker
+	subscribeWorker *subscribeWorker
 }
 
 func NewReplicationApiService(
@@ -36,16 +39,22 @@ func NewReplicationApiService(
 	registrant *registrant.Registrant,
 	store *sql.DB,
 ) (*Service, error) {
-	worker, err := StartPublishWorker(ctx, log, registrant, store)
+	publishWorker, err := startPublishWorker(ctx, log, registrant, store)
 	if err != nil {
 		return nil, err
 	}
+	subscribeWorker, err := startSubscribeWorker(ctx, log, store)
+	if err != nil {
+		return nil, err
+	}
+
 	return &Service{
-		ctx:        ctx,
-		log:        log,
-		registrant: registrant,
-		store:      store,
-		worker:     worker,
+		ctx:             ctx,
+		log:             log,
+		registrant:      registrant,
+		store:           store,
+		publishWorker:   publishWorker,
+		subscribeWorker: subscribeWorker,
 	}, nil
 }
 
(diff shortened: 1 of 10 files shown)


richardhuaaa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/160?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#160** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/160?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#159** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/159?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

pkg/api/subscribeWorker.go
```diff
@@ -0,0 +1,196 @@
...
+// and may be dropped if full
+type subscribeWorker struct {
+	ctx context.Context
+	log *zap.Logger
+
+	dbSubscription <-chan []queries.GatewayEnvelope
+	// Assumption: listeners cannot be in multiple slices
+	globalListeners     []listener
+	originatorListeners map[uint32][]listener
+	topicListeners      map[string][]listener
```
richardhuaaa, line 40: This is a little different from the previous implementation. Storing it in this format allows the subscribeWorker to only iterate over relevant listeners for each payload, rather than iterating over all listeners.

However, for doing listener cleanup, I think there will need to be additional metadata on listeners in the reverse direction (storing the topics/originators for each listener, so that it can be cleaned from the relevant maps)

pkg/api/service.go
```diff
@@ -55,15 +64,63 @@ func (s *Service) Close() {
...
+					err := stream.Send(&message_api.BatchSubscribeEnvelopesResponse{
+						Envelopes: envs,
+					})
+					if err != nil {
+						log.Error("sending envelope to subscribe", zap.Error(err))
+					}
+				}()
+			} else {
+				// TODO(rich) Recover from backpressure
+				log.Info("stream closed due to backpressure")
```
richardhuaaa, line 105: With cursors, recovery from backpressure should be possible. Basically we drain the channel, and then restart the subscription from scratch using the cursor. That in turn will listen on a new channel, but pull from the DB first (ignoring duplicates coming in on the channel).

pkg/api/subscribeWorker.go
```diff
@@ -0,0 +1,196 @@
...
+	requests []*message_api.BatchSubscribeEnvelopesRequest_SubscribeEnvelopesRequest,
+) (<-chan []*message_api.OriginatorEnvelope, error) {
+	subscribeAll := false
+	topics := make(map[string]bool, len(requests))
+	originators := make(map[uint32]bool, len(requests))
+
+	if len(requests) > maxSubscriptionsPerClient {
+		// When a client subscribes to too many originators or topics, we treat it as a request to
+		// subscribe to all instead of throwing an error. We rely on the client's existing
+		// filtering logic rather than forcing clients to respond to an error.
```
richardhuaaa, line 152: I believe this is different from the old implementation. It means that we can set `maxSubscriptionsPerClient` lower and arbitrarily modify it, and use less memory/iterate through fewer subscriptions.

I suspect this is a better behavior for clients to handle, because clients need to have filtering logic anyway. It does mean that clients have to have the ability to discard payloads that are not relevant to them.

pkg/api/service.go
```diff
@@ -55,15 +64,63 @@ func (s *Service) Close() {
...
+				exit = true
+			}
+		case <-stream.Context().Done():
+			log.Debug("stream closed")
+			exit = true
```
neekolas, line 110: You could also do this with a [labeled loop](https://www.ardanlabs.com/blog/2013/11/label-breaks-in-go.html) and a break

pkg/api/service.go
```diff
@@ -55,15 +64,63 @@ func (s *Service) Close() {
...
+		}
+	}
+	return nil
 }
 
 func (s *Service) QueryEnvelopes(
 	ctx context.Context,
 	req *message_api.QueryEnvelopesRequest,
 ) (*message_api.QueryEnvelopesResponse, error) {
+	log := s.log.Named("query")
```
neekolas, line 123: Instead of taking over the name of the logger, it might be more readable to have the struct hold a logger named `api` and to add context for each method using `With` for each endpoint. Then we can group together all the API handlers in DataDog.

pkg/api/subscribeWorker.go
neekolas: Feels a little odd to change the meaning of the request silently. At the very least we should log it, but it might be more intuitive to just return an error and have the caller resubscribe to all. Or, since we control the client, just have it be smart enough to create the right subscription based on the limit the first time.

pkg/db/queries.sql
```diff
@@ -66,3 +66,12 @@ LIMIT @num_rows;
 DELETE FROM staged_originator_envelopes
 WHERE id = @id;
 
+-- name: SelectVectorClock :many
+SELECT
+	originator_node_id,
+	max(originator_sequence_id)::BIGINT AS originator_sequence_id
```
neekolas, line 72: With the right indexes this query should be fast. We'll want to do some `EXPLAIN ANALYZE`ing on it at some point to make sure we are taking advantage of the indexes here.

neekolas (approved):

pkg/api/service.go
```diff
@@ -55,15 +64,63 @@ func (s *Service) Close() {
...
+	if len(requests) == 0 {
+		return status.Errorf(codes.InvalidArgument, "missing requests")
+	}
+
+	ch, err := s.subscribeWorker.addListeners(requests)
+	if err != nil {
+		return status.Errorf(codes.InvalidArgument, "invalid subscription request: %v", err)
+	}
+
+	var streamLock sync.Mutex
```
mkysel, line 88: what is this lock protecting? Who is racing against who here?

pkg/api/subscribeWorker.go
```diff
@@ -0,0 +1,196 @@
...
+			return
+		case new_batch := <-s.dbSubscription:
+			for _, row := range new_batch {
+				s.dispatch(&row)
+			}
+		}
+	}
+}
+
+func (s *subscribeWorker) dispatch(
```
mkysel, line 112: can a client subscribe to both an originator and a topic? I assume that each message comes from exactly 1 originator and belongs to exactly 1 topic. Does this mean that a client might receive each message twice if it subscribes to both?

pkg/api/subscribeWorker.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	ch := make(chan []*message_api.OriginatorEnvelope, subscriptionBufferSize)
+
+	if subscribeAll {
+		if len(topics) > 0 || len(originators) > 0 {
+			return nil, fmt.Errorf("cannot filter by topic or originator when subscribing to all")
+		}
+		// TODO(rich) thread safety
+		s.globalListeners = append(s.globalListeners, ch)
```
mkysel, line 181: should becoming an globalListener remove said client from all topic/origin listeners?

mkysel:
if a client installation subscribes multiple times, those subscriptions are all separate?

pkg/api/subscribeWorker.go
```diff
@@ -0,0 +1,196 @@
...
+
+	return worker, nil
+}
+
+func (s *subscribeWorker) start() {
+	for {
+		select {
+		case <-s.ctx.Done():
+			return
+		case new_batch := <-s.dbSubscription:
```
mkysel, line 104: I am a bit uncomfortable that we are passing a slice to a channel. In this simple use case its correct, but slices are mutable and not thread safe. So it makes me not trust this code.

pkg/api/service.go
mkysel: would you explain the backpressure logic to me next time we chat?

pkg/api/subscribeWorker.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	for _, listener := range s.globalListeners {
+		select {
+		case listener <- []*message_api.OriginatorEnvelope{env}:
+		default:
+		}
+	}
+}
+
+func (s *subscribeWorker) addListeners(
```
mkysel, line 142: can I be a bit nitpicky here?

addListeners sounds like we are passing in listeners as parameters and adding them to the worker. The fact that we get back a channel that we should listen to makes it slightly worse.

I am the worst when it comes to naming, but I would have named it something like `createChannel` or something.

pkg/api/subscribeWorker.go
```diff
@@ -0,0 +1,196 @@
...
+	bytes := row.OriginatorEnvelope
+	env := &message_api.OriginatorEnvelope{}
+	err := proto.Unmarshal(bytes, env)
+	if err != nil {
+		s.log.Error("Failed to unmarshal envelope", zap.Error(err))
+		return
+	}
+	for _, listener := range s.originatorListeners[uint32(row.OriginatorNodeID)] {
+		select {
+		case listener <- []*message_api.OriginatorEnvelope{env}:
```
mkysel, line 124: making a comment on this line, but it applies to all.

This feels like a very easy denial of service attack. If any channel/subscriber can't keep up, none will receive messages.

Am I reading the code correctly?

mkysel:
I believe that you say this in the description of the PR, but I just want to confirm. Right now the client `lastSeen` is ignored. Implementing that is represented by `Subscribing from a vector clock cursor`?

pkg/api/service.go
mkysel: or just explicitly return nil such as:
```

To get rid of the exit variable and simplify the loop termination, you can use a for loop combined with a return or break statement inside the select block. This approach avoids the need for an explicit exit variable and results in a cleaner code structure. Here's how you can rewrite the code:

Refactored Code:
go
Copy code
for {
    select {
    case envs, open := <-ch:
        if open {
            func() {
                streamLock.Lock()
                defer streamLock.Unlock()
                err := stream.Send(&message_api.BatchSubscribeEnvelopesResponse{
                    Envelopes: envs,
                })
                if err != nil {
                    log.Error("sending envelope to subscribe", zap.Error(err))
                }
            }()
        } else {
            // TODO: Recover from backpressure
            log.Info("stream closed due to backpressure")
            return nil  // Terminate the function when the channel is closed
        }

    case <-stream.Context().Done():
        log.Debug("stream closed")
        return nil  // Exit the function when the stream is done

    case <-s.ctx.Done():
        log.Info("service closed")
        return nil  // Exit when the service context is done
    }
}```

pkg/api/subscribeWorker.go
richardhuaaa: Can you be more specific about the case you are worried about? In general I'm not sure if there's a hard-and-fast rule that all objects passed into a channel must be immutable (although it would be suspect if they were mutated while they were in the channel, which we don't do)

pkg/api/service.go
richardhuaaa: Will do!

pkg/api/subscribeWorker.go
richardhuaaa: No, this is documented in the struct itself as well as guarded against below. In general, clients will be subscribing by topic only, and replication servers will be subscribing by originator only, so there's no need to have both in the same request

pkg/api/subscribeWorker.go
richardhuaaa: The `select{}` with `default` is a non-blocking write to the buffered channel, so we don't depend on subscribers to keep up

pkg/api/subscribeWorker.go
richardhuaaa: Yes I agree, I was thinking of renaming this to `listen(requests)`

pkg/api/subscribeWorker.go
richardhuaaa: The check on L177 means that the client won't have any topic/origin listeners if they are subscribing globally

pkg/api/service.go
richardhuaaa: Great question. The loop below actually came from the implementation on the old server. The lock is used for the duration of the send operation on the stream, but I actually don't see any need for it because the loop is running synchronously. Maybe it was some kind of historical cruft - I'll remove it tomorrow as well as make the style changes we're requesting.

richardhuaaa:
> if a client installation subscribes multiple times, those subscriptions are all separate?

For each `batchSubscribe()` call, the client gets a separate stream. Within a single `batchSubscribe()` call, there may be multiple requests, all of which get aggregated into a single stream.

> I believe that you say this in the description of the PR, but I just want to confirm. Right now the client lastSeen is ignored. Implementing that is represented by Subscribing from a vector clock cursor?

Yep, that's right. Picking up historical messages from before the stream was requested (without dropping messages or having duplicates) is a bit more complicated, so I'm splitting it off into a separate PR so as not to make this one harder to review.

pkg/api/subscribeWorker.go
mkysel: Gotcha. So as the code is written in the PR; if the channel is full we don't broadcast more. This will lead to gaps I think. The eventual plan is to kill listeners that are not keeping up?

pkg/api/subscribeWorker.go
mkysel: I was assuming incorrectly that there is only one channel per client and we can merge these all together.

pkg/api/subscribeWorker.go
richardhuaaa: I'm just thinking through the client logic, and think it might be simpler this way?

Option 1: List in the docs that at an unspecified number of batched requests, the stream will subscribe to all updates. Clients should always be filtering the stream by topic regardless (esp if they subscribe to multiple topics), the only addition is they should drop any irrelevant payloads.

Option 2: List in the docs a specific max number of batched requests. Clients need to track the number of batched requests they issue, and switch to a subscribeAll subscription at that number (including if they add requests via the bidirectional stream, they would need to tear it down and re-establish). Once at that number, they need to implement the logic to filter the stream and drop irrelevant payloads regardless.

WDYT? I don't have a strong opinion, so happy to switch to an error if you still feel it is unintuitive.

pkg/db/queries.sql
richardhuaaa: The primary key on this table is `originator_node_id, originator_sequence_id`, which I think is a b-tree index, so hopefully it'll be fast - but would love to get some analyzing done

richardhuaaa:
* **#166** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/166?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#160** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/160?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

richardhuaaa:
* **#166** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/166?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#160** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/160?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pkg/api/subscribeWorker.go
neekolas: Are we imagining that this behaviour is going to be extended to clients? Like at some point my phone might get the firehose of all messages on XMTP? A public figure user is going to subscribe to thousands of topics since they have thousands of conversations...but that's still going to be a very different throughput to getting every message on XMTP since most of their subscriptions would be dormant.

neekolas: For topic subscriptions, we should be expecting that users can be subscribed to tens of thousands of topics on the client. We can maybe adjust the `subscribe_all_messages` behaviour on the client to only subscribe to consented conversations once we have consent native to V3, which'll lower the number to something reasonable. But still could be thousands.

Bots could have hundreds of thousands of convos and might at least have the compute to handle the firehose. But it's a lot more outbound bandwidth for us to stream all messages than it is to listen to 100k topics that are mostly idle.

richardhuaaa: After our chat, the thing I misunderstood is that the client will always respond to an error by subscribing to all, which is not true - a client may choose to subscribe to fewer topics, or split them across multiple subscriptions. Changed to an error!

PR #163, opened 2024-09-16
Refactor unit tests

richardhuaaa:
- Split up query, publish and (not yet added) subscribe tests
- Test the server through the GRPC client rather than calling server methods directly (needed to test streams properly) 
- Fix up server and DB cleanup after unit tests - this also makes test logs much easier to troubleshoot


richardhuaaa:
* **#163** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/163?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pkg/testutils/api.go
```diff
@@ -0,0 +1,56 @@
...
+	}, nil)
+	registrant, err := registrant.NewRegistrant(ctx, queries.New(db), mockRegistry, privKeyStr)
+	require.NoError(t, err)
+
+	svr, err := api.NewAPIServer(ctx, db, log, 0 /*port*/, registrant, true /*enableReflection*/)
+	require.NoError(t, err)
+
+	return svr, db, func() {
+		log.Info("-------- Cleaning up server and DB ----------")
+		cancel()
```
richardhuaaa, line 38: By canceling the context, we clean up dependencies such as the subscribe and publish worker, so that they will stop querying the DB and returning errors.

I'm not sure if going forward, we should keep public `Close()` methods, or rely on cleanup being triggered via closing the context.

neekolas: If the cleanup is just "stop this thing" it's fine (and preferable) to do it through the context. 

The one advantage to having a close method is waiting for cleanup tasks that take time. Calling `Close` as part of shutdown can make sure that the `Close` method has completed before the server fully terminates. The server may shut down before all listeners to the `context.Done()` channel have picked up the message.

neekolas (approved):

PR #166, opened 2024-09-18
Thread-safe subscriptions and listener cleanup

richardhuaaa:
This PR adds:
- Thread-safety for mutation of listener maps
- Safe cleanup flow for listener channels

As a general principle, we use Golang's `sync.map`. This is an optimistic concurrency pattern that restricts contention to a per-key level, and separates reads and writes via a read-only map and dirty map. So the dispatch loop is not affected - any mutations to the sync maps are performed outside of the dispatching goroutine.

I've also added a `RWMutex` to synchronize between adding and removing listeners. The main thing we are protecting against is that when removing a listener, we may want to delete the `listenerset` if it is empty - but we can't perform the emptiness check and the deletion step atomically without a mutex. I think this should be okay, because in the current server we are getting something like 10 new subscriptions a second.

Would love to do some benchmarking on this later to make sure we've made the right tradeoffs, I can see us changing out the implementation underneath depending on where the bottleneck is.

Closes https://github.com/xmtp/xmtpd/issues/125

### DIFF
--- a/pkg/api/subscribeWorker.go
+++ b/pkg/api/subscribeWorker.go
@@ -44,7 +44,7 @@ func (ls *listenerSet) removeListener(l *listener) {
 	ls.Delete(l)
 }
 
-func (ls *listenerSet) isEmpty() bool {
+func (ls *listenerSet) IsEmpty() bool {
 	empty := true
 	ls.Range(func(_, _ interface{}) bool {
 		empty = false
@@ -65,20 +65,13 @@ func (lm *listenersMap[K]) addListener(key K, l *listener) {
 }
 
 func (lm *listenersMap[K]) removeListener(key K, l *listener) {
-	for {
-		value, ok := lm.Load(key)
-		if !ok || value == nil {
-			return // Key doesn't exist, nothing to do
-		}
-		set := value.(*listenerSet)
-		set.removeListener(l)
-
-		if !set.isEmpty() || lm.CompareAndDelete(key, value) {
-			return
-		}
-		// Another goroutine either removed the key already or added a listener,
-		// try again. Should only retry once at most.
+	value, ok := lm.Load(key)
+	if !ok || value == nil {
+		return // Key doesn't exist, nothing to do
 	}
+	set := value.(*listenerSet)
+	set.removeListener(l)
+	// TODO(rich): Delete keys that hold empty sets
 }
 
 // A worker that listens for new envelopes in the DB and sends them to subscribers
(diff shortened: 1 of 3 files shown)


richardhuaaa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/166?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#166** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/166?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#160** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/160?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

pkg/api/subscribeWorker.go
```diff
@@ -24,7 +23,56 @@ const (
...
+}
+
+func (lm *listenersMap[K]) removeListener(key K, l *listener) {
+	value, ok := lm.Load(key)
+	if !ok || value == nil {
+		return // Key doesn't exist, nothing to do
+	}
+	set := value.(*listenerSet)
+	set.removeListener(l)
+	// TODO(rich): Delete keys that hold empty sets
```
richardhuaaa, line 74: The only thing I haven't been able to solve with this approach is here. I haven't worked out a way to delete keys for which there are no more listeners, in a way that is thread-safe and doesn't compromise readability or performance.

The issue is that checking that the set is empty, and then deleting it, is not an atomic operation. And I suspect that using something like `sync.map.CompareAndDelete()` won't work, because that is likely doing pointer comparison.

Any suggestions are welcome, I can also keep thinking about this - one approach would be a granular map of RWMutex, another approach would be to stop storing these listener lookups and iterate through all listeners on every payload. I'm assuming this has to be solved, otherwise we end up holding all topics in memory on the server which I assume is too much.

mkysel:
On a high level this looks good. I won't have the time this week to look into it deeper. Feel free to move on without me.

pkg/api/subscribeWorker.go
richardhuaaa: I'm leaning towards using a channel here. Moving PR back to draft for now while I investigate that approach.

neekolas (approved):

PR #170, opened 2024-09-20
Subscribe from cursor

richardhuaaa:
This adds the ability to stream historical payloads starting from a cursor, and seamlessly transition to streaming new incoming payloads once caught up.

I expect there to be overlap in payloads between the catch-up process and the newly incoming payloads - to handle this, we simply update the vector clock as we send each payload into the stream, and discard incoming payloads that are older than our vector clock.

https://github.com/xmtp/xmtpd/issues/126


richardhuaaa:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/170?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#170** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/170?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#166** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/166?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#160** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/160?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

pkg/api/server.go
```diff
@@ -44,7 +44,7 @@ func NewAPIServer(
 	enableReflection bool,
 	messagePublisher blockchain.IBlockchainPublisher,
 ) (*ApiServer, error) {
-	grpcListener, err := net.Listen("tcp", fmt.Sprintf("0.0.0.0:%d", port))
+	grpcListener, err := net.Listen("tcp", fmt.Sprintf("localhost:%d", port))
```
richardhuaaa, line 47: This gets rid of those pesky 'Allow incoming connections' dialogue when running unit tests - we only want to listen to incoming connections on localhost

pkg/api/service.go
```diff
@@ -91,18 +164,23 @@ func (s *Service) SubscribeEnvelopes(
 	}
 
 	ch := s.subscribeWorker.listen(stream.Context(), query)
+	err = s.catchUpFromCursor(stream, query, log)
```
richardhuaaa, line 167: I spent a lot of time thinking about whether there is a race condition here that would cause a 'gap' in payloads between when the catch-up ends and when the next batch of envelopes comes in from the subscribe worker. I *think* there is no race condition, here is my reasoning:

```
SUBSCRIBE WORKER ALGORITHM
- In a loop:
   - Fetch new payloads from DB
   - Fetch listeners
   - Push to listener channels
   - Wait 100ms
```

```
SUBSCRIBE REQUEST ALGORITHM (this function)
- Add self to subscribe worker listeners list
- Catch up from DB
- Pull from the listener channel
```

In order for there to be a race condition, the subscribe worker needs to fetch from the DB *after* the subscribe request's catch-up, but *not* push to the subscribe request's listener. I think that's not possible, because the subscribe request's listener was added to the subscribe worker's listener list before the catch-up was started.

pkg/api/service.go
```diff
@@ -91,18 +164,23 @@ func (s *Service) SubscribeEnvelopes(
...
-					Envelopes: envs,
-				})
+				err := s.sendEnvelopes(stream, query, envs)
 				if err != nil {
 					return status.Errorf(codes.Internal, "error sending envelope: %v", err)
 				}
 			} else {
-				// TODO(rich) Recover from backpressure
+				// TODO(rich) Reset whole subscribe flow from new cursor
 				log.Debug("channel closed by worker")
```
richardhuaaa, line 184: If the channel gets filled up, I think handling it is fairly simple:

1. Drain the channel, updating the cursor as we do so.
2. Restart the whole subscription request starting from that cursor (setting up a new listener, and catching up from that cursor).

If the consumer on the stream is slow, then it just means that the stream will be permanently behind the database, and the consumer will be in the catch-up flow forever. The catch-up flow simply does a poll every 50 ms, blocking while it sends payloads down before the next poll.

pkg/api/service.go
```diff
@@ -72,6 +73,78 @@ func (s *Service) Close() {
...
+	for _, env := range envs {
+		if cursor[uint32(env.OriginatorNodeID())] >= env.OriginatorSequenceID() {
+			continue
+		}
+
+		// TODO(rich): Either batch send envelopes, or modify stream proto to
+		// send one envelope at a time.
+		err := stream.Send(&message_api.SubscribeEnvelopesResponse{
+			Envelopes: []*message_api.OriginatorEnvelope{env.Proto()},
+		})
```
richardhuaaa, line 139: I can restructure this loop to batch send, I'm just unsure if that is more performant or not. We would be trading off the time to copy the envelopes to another array, versus the efficiency gain of batch sending. I'm guessing the efficiency gain will be greater, but I'm really unsure

pkg/api/message/service.go
```diff
@@ -117,19 +122,104 @@ func (s *Service) SubscribeEnvelopes(
...
+		err = s.sendEnvelopes(stream, query, envs)
+		if err != nil {
+			return status.Errorf(codes.Internal, "error sending envelopes: %v", err)
+		}
+		if len(rows) < int(maxRequestedRows) {
+			// There were no more envelopes in DB at time of fetch
+			break
+		}
+		// TODO(rich): Determine default rate limits and set query interval to match
+		time.Sleep(50 * time.Millisecond)
```
richardhuaaa, line 172: Would love for us to work out what rate limits we want to impose by default (e.g. if a client sends many requests), and what setting to use here that would fall within that rate limit (which is the equivalent of requesting every 50ms)

neekolas: The actual bookkeeping of the rate limits is going to be interesting here, since a query that involves catchup could yield 1 database query or >10,000.

I wonder if we are going to have to meter based on the actual number of underlying database queries vs some forward projection of how many queries the request is going to take. Like we would decrement your rate limit on every page of results pulled.

What about backwards from a target of 10 requests/second as a generous but not extravagant starting place?

pkg/api/message/service.go
```diff
@@ -117,19 +122,104 @@ func (s *Service) SubscribeEnvelopes(
...
+		query.LastSeen = &envelopesProto.VectorClock{
+			NodeIdToSequenceId: cursor,
+		}
+	}
+	for _, env := range envs {
+		if cursor[uint32(env.OriginatorNodeID())] >= env.OriginatorSequenceID() {
+			continue
+		}
+
+		// TODO(rich): Either batch send envelopes, or modify stream proto to
```
neekolas, line 194: WDYT?

I like the idea of batching, since it lets clients update their bookkeeping of the lastSeen cursor once per batch instead of once per envelope.

pkg/api/server.go
neekolas: Interesting...

Great find

pkg/envelopes/originator.go
```diff
@@ -12,6 +12,14 @@ type OriginatorEnvelope struct {
 	UnsignedOriginatorEnvelope UnsignedOriginatorEnvelope
 }
 
+func NewOriginatorEnvelopeFromBytes(bytes []byte) (*OriginatorEnvelope, error) {
```
neekolas, line 15: This is going to cause conflicts. I wound up implementing the same thing in another PR that is about to be merged in.

Sorry about that.

neekolas (approved):

pkg/api/message/service.go
richardhuaaa: Sounds good - updated to a 100ms paging interval for now with a 10 req/sec target in mind. We may need to adjust this on a more granular level for firehose-type applications, like node<>node replication. Will follow up on that when we are reading the JWT and poking a hole in the default rate limit

pkg/api/message/service.go
richardhuaaa: Sounds good - will add to next PR

PR #171, opened 2024-09-20
Run smart contract registry locally

neekolas:

PR #175, opened 2024-09-25
Publish IdentityUpdates to blockchain and read back

neekolas:
## tl;dr

- Publishes IdentityUpdates to blockchain as part of the publsh API
- Fixes some bugs in the `IdentityUpdate` storing flow
- Prevents duplicate `IdentityUpdate`s from being saved


neekolas:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/175?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#176** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/176?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#175** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/175?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#171** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/171?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

contracts/src/Nodes.sol
```diff
@@ -133,11 +133,14 @@ contract Nodes is ERC721, Ownable {
     Get all nodes regardless of their health status
      */
     function allNodes() public view returns (NodeWithId[] memory) {
-        uint32 totalNodeCount = _nodeIdCounter;
+        uint32 totalNodeCount = _nodeIdCounter - _startingNodeId;
```
richardhuaaa, line 136: Could we also increment the Node ID by 100 with every new node? Leaves some flexibility in the future for any design changes where an organization runs multiple nodes/sub-nodes (or rotates their keys, etc). If this contract is being re-written anyway I suppose it doesn't matter if we do it now

pkg/api/service.go
```diff
@@ -24,19 +27,22 @@ const (
...
-	store           *sql.DB
-	publishWorker   *publishWorker
-	subscribeWorker *subscribeWorker
+	ctx              context.Context
+	log              *zap.Logger
+	registrant       *registrant.Registrant
+	store            *sql.DB
+	publishWorker    *publishWorker
+	subscribeWorker  *subscribeWorker
+	messagePublisher blockchain.IMessagePublisher
```
richardhuaaa, line 36: Could we name this something like `blockchainPublisher` to be clear, here and in `server.go`? There are multiple places publishes can go, including into the DB.

pkg/api/service.go
```diff
@@ -225,12 +238,89 @@ func (s *Service) PublishEnvelope(
...
+		ctx,
+		inboxId,
+		identityUpdateBytes,
+	)
+}
+
+func (s *Service) GetInboxIds(
+	ctx context.Context,
+	req *message_api.GetInboxIdsRequest,
+) (*message_api.GetInboxIdsResponse, error) {
```
richardhuaaa, line 283: How would you feel about us replacing this with a `batchQuery` endpoint eventually? I'm also curious if for the demo the client could have done the batching internally with multiple requests

pkg/blockchain/interface.go
```diff
@@ -47,3 +43,8 @@ type NodeRegistry interface {
 		httpAddress string,
 	) error
 }
+
+type IMessagePublisher interface {
+	PublishIdentityUpdate(ctx context.Context, inboxId [32]byte, identityUpdate []byte) error
+	PublishGroupMessage(ctx context.Context, groupdId [32]byte, message []byte) error
```
richardhuaaa, line 49: PublishCommit?

pkg/db/queries.sql
```diff
@@ -75,3 +75,56 @@ FROM
...
+		address,
+		inbox_id);
+
+-- name: GetLatestSequenceId :one
+SELECT
+	COALESCE(max(originator_sequence_id), 0)::BIGINT AS originator_sequence_id
+FROM
+	gateway_envelopes
+WHERE
+	originator_node_id = @originator_node_id;
```
richardhuaaa, line 129: We originally added this as a hack, but I guess we can keep it permanently? It does make the server more robust against edge cases

pkg/indexer/storer/identityUpdate.go
```diff
@@ -35,21 +50,78 @@ func (s *IdentityUpdateStorer) StoreLog(ctx context.Context, event types.Log) Lo
 		return NewLogStorageError(err, false)
 	}
 
-	// TODO:nm figure out topic structure
+	latestSequenceId, err := s.queries.GetLatestSequenceId(ctx, IDENTITY_UPDATE_ORIGINATOR_ID)
+	if err != nil {
+		return NewLogStorageError(err, true)
+	}
+	if uint64(latestSequenceId) >= msgSent.SequenceId {
```
richardhuaaa, line 57: Should we log a warning or error here? I assume in the future we will have a mechanism to start syncing from the correct location?

pkg/db/queries.sql
```diff
@@ -75,3 +75,56 @@ FROM
...
+		inbox_id,
+		MAX(association_sequence_id)
+	FROM
+		address_log AS a
+	WHERE
+		a.address = @address
+		AND a.inbox_id = decode(@inbox_id, 'hex')
+	GROUP BY
+		address,
+		inbox_id);
```
richardhuaaa, line 121: Do we need this sub-query, given `(address, inbox_id)` is a primary key that uniquely identifies the row? The PK is a small simplifying change compared to the old server

pkg/db/queries.sql
```diff
@@ -75,3 +75,56 @@ FROM
...
+			AND revocation_sequence_id IS NULL
+		GROUP BY
+			address) b ON a.address = b.address
+	AND a.association_sequence_id = b.max_association_sequence_id;
+
+-- name: InsertAddressLog :one
+INSERT INTO address_log(address, inbox_id, association_sequence_id, revocation_sequence_id)
+	VALUES (@address, decode(@inbox_id, 'hex'), @association_sequence_id, @revocation_sequence_id)
+RETURNING
+	*;
```
richardhuaaa, line 102: Could we follow `idempotent inserts` from https://github.com/xmtp/xmtpd/pull/113, so that all inserts have consistent behavior? I think that also gives us a way to address the scenario where this row already exists

pkg/utils/hex.go
```diff
@@ -0,0 +1,7 @@
+package utils
+
+import "encoding/hex"
+
+func HexEncode(data []byte) string {
+	return hex.EncodeToString(data)
+}
```
richardhuaaa, line 7: Should we just move the `hex.go` from `/testutils/` here?

richardhuaaa (commented):
This PR is a big one! I have a few nits on first pass, happy to look at it again quickly

pkg/blockchain/interface.go
neekolas: Technically it doesn't _have_ to be a commit. If you really really want censorship resistance you could just publish everything here

pkg/db/queries.sql
neekolas: Even once we have checks in place to start the blockchain indexing from the last seen block, there might be partially completed blocks from a past session. I'm OK with leaving for now.

pkg/indexer/storer/identityUpdate.go
neekolas: Going to add that mechanism next.

pkg/db/queries.sql
neekolas: Not sure I follow the suggestion. IIRC we need the subquery because an address can be associated with multiple inbox IDs, and we need to find the one with the highest `association_sequence_id`?

Maybe you can sketch out what you have in mind as an alternate.

pkg/db/queries.sql
neekolas: Done. We don't even look at the returned value, so I also dropped the `RETURNING` clause

pkg/utils/hex.go
neekolas: Nice catch

pkg/api/service.go
neekolas: It would have to be its own`batchQuery` endpoint, since the address log is separate from the gateway envelopes table.

contracts/src/Nodes.sol
neekolas: Think I might do this as a separate PR. This thing is too big as it is

pkg/db/queries.sql
richardhuaaa: I totally misread what this was doing, my bad!

pkg/db/queries.sql
```diff
@@ -75,3 +75,46 @@ FROM
...
+		FROM
+			address_log
+		WHERE
+			address = ANY (@addresses::TEXT[])
+			AND revocation_sequence_id IS NULL
+		GROUP BY
+			address) b ON a.address = b.address
+	AND a.association_sequence_id = b.max_association_sequence_id;
+
+-- name: InsertAddressLog :exec
```
richardhuaaa, line 98: @neekolas - Very nitpicky, but for idempotent inserts I think it's always better to use `:execrows` if we're not using `:one`, even if we ignore the result in go code via `_, err = `, because it makes the intent explicit to readers (we want to do nothing if the row already exists).

In this case though, because we have a primary key on `address, inbox_id` (a change from the old code), it means that an address is being added to an inbox_id but a row already exists for the `(address, inbox_id)` pair -
- If the row is revoked with a lower sequence ID, we should replace the row
- If the row is unrevoked and has a higher sequence ID, we should either surface an error or silently return success

Could probably change the query to do something like `... ON CONFLICT WHERE revocation_sequence_id IS NULL OR revocation_sequence_id < @association_sequence_id AND association_sequence_id < @association-sequence_id DO UPDATE ...`

richardhuaaa (approved):

pkg/db/queries.sql
neekolas: This is a very good point. In the fog of last week's hackathon I forgot that we changed the schema, so I left the query pretty much intact. Will push up a change to update the query

PR #206, opened 2024-10-08
Add sync service and worker

richardhuaaa:
This adds a sync service that subscribes to every other originator, based on the pair programming @mkysel and I did two weeks ago. For now, tests are performed on localhost with one server on each port. This is the most basic version, future PR's will add missing pieces such as handling originator outages, originator payload validation, server configuration, etc.

A bunch of auto-generated proto changes also got mixed into this PR. How do we feel about removing `dev/generate` from `dev/up`?

Closes https://github.com/xmtp/xmtpd/issues/181


neekolas:
I feel good about removing dev/generate from up. We have them as separate commands in node go for exactly this reason.

pkg/sync/syncWorker.go
```diff
@@ -0,0 +1,205 @@
...
+	return s, nil
+}
+
+func (s *syncWorker) start() error {
+	nodes, err := s.nodeRegistry.GetNodes()
+	if err != nil {
+		return err
+	}
+	for _, node := range nodes {
+		if node.NodeID == s.registrant.NodeID() || !node.IsHealthy {
```
mkysel, line 58: you also want to skip `NodeID == 0` and `!node.IsValid`

pkg/sync/syncWorker.go
```diff
@@ -0,0 +1,205 @@
...
+	tracing.GoPanicWrap(
+		s.ctx,
+		&s.wg,
+		fmt.Sprintf("node-subscribe-%d", node.NodeID),
+		func(ctx context.Context) {
+			var err error
+			for {
+				if err != nil {
+					log.Error(fmt.Sprintf("Error: %v, retrying...", err))
+					time.Sleep(1 * time.Second)
```
mkysel, line 80: this should eventually be done by an exponential backoff

pkg/sync/syncWorker.go
```diff
@@ -0,0 +1,205 @@
...
+		},
+	)
+}
+
+func (s *syncWorker) connectToNode(node registry.Node) (*grpc.ClientConn, error) {
+	addr := node.HttpAddress
+	log.Info(fmt.Sprintf("attempting to connect to %s", addr))
+	conn, err := grpc.NewClient(
+		addr,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
```
mkysel, line 106: Will this work with TLS in the real environment?

pkg/sync/syncWorker.go
```diff
@@ -0,0 +1,205 @@
...
+	stream, err := client.BatchSubscribeEnvelopes(
+		s.ctx,
+		&message_api.BatchSubscribeEnvelopesRequest{
+			Requests: []*message_api.BatchSubscribeEnvelopesRequest_SubscribeEnvelopesRequest{
+				{
+					Query: &message_api.EnvelopesQuery{
+						Filter: &message_api.EnvelopesQuery_OriginatorNodeId{
+							OriginatorNodeId: node.NodeID,
+						},
+						LastSeen: nil,
```
mkysel, line 127: is this another `todo`?

pkg/sync/syncWorker.go
```diff
@@ -0,0 +1,205 @@
...
+			),
+			Topic:              clientEnvelope.GetAad().GetTargetTopic(),
+			OriginatorEnvelope: originatorBytes,
+		},
+	)
+	if err != nil {
+		log.Error("Failed to insert gateway envelope", zap.Error(err))
+	} else if inserted == 0 {
+		// Envelope was already inserted by another worker
+		log.Error("Envelope already inserted")
```
mkysel, line 203: can we log this as `Debug`? Given that we have 2+ workers, it will be a very common scenario.

mkysel:
All the protos stuff should be handled by https://github.com/xmtp/xmtpd/pull/207. That will hopefully simplify this PR and only review changes related to the new feature.

pkg/sync/syncWorker.go
```diff
@@ -0,0 +1,205 @@
...
+func startSyncWorker(
+	ctx context.Context,
+	log *zap.Logger,
+	nodeRegistry registry.NodeRegistry,
+	registrant *registrant.Registrant,
+	store *sql.DB,
+) (*syncWorker, error) {
+	s := &syncWorker{
+		ctx:          ctx,
+		log:          log.With(zap.String("method", "syncWorker")),
```
neekolas, line 40: I think we should reserve the key `method` for actual APIs. For this, `log.Named("syncWorker")` would be more appropriate.

pkg/sync/syncWorker.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+		for _, env := range envs.Envelopes {
+			s.insertEnvelope(env)
+		}
+	}
+}
+
+func (s *syncWorker) insertEnvelope(env *message_api.OriginatorEnvelope) {
+	log.Info(fmt.Sprintf("Replication server received envelope %s", env))
+	// TODO(rich) Validation logic - share code with API service and publish worker
```
neekolas, line 159: I've been meaning to make some nice utility functions and wrappers for these anyways. I can take that TODO

pkg/sync/syncWorker.go
neekolas: This brings up a good point. We don't have a standard way to specify if the server expects a TLS connection or not. In deployed environments we want to make sure that all servers connect via TLS, but it's a pain to enforce that locally.

We could assume that urls prefixed with `https://` are TLS and anything else isn't. Or we could add more metadata to the node registry.

pkg/sync/syncWorker.go
```diff
@@ -0,0 +1,205 @@
...
+		func(ctx context.Context) {
+			var err error
+			for {
+				if err != nil {
+					log.Error(fmt.Sprintf("Error: %v, retrying...", err))
+					time.Sleep(1 * time.Second)
+				}
+				err = nil
+
+				conn, err := s.connectToNode(node)
```
neekolas, line 84: We should log these errors

pkg/sync/syncWorker.go
richardhuaaa: I don't think the node registry will ever have NodeID < 100, so we shouldn't need to check for it - that was just a workaround we had earlier.

Added a check for `node.IsValidConfig`, that seems to be new - @neekolas why is this needed? Would it make sense to collapse it into `node.IsHealthy`?

pkg/sync/syncWorker.go
richardhuaaa: Agree, will work on that as part of handling node failures

pkg/sync/syncWorker.go
richardhuaaa: Great catch, I added a work item `Allow registry to specify TLS endpoints for prod and non-TLS endpoints for dev` for now. We should also figure out what format we want the URL's to be in - right now we're setting them to a GRPC-specific format (prefixed with `dns://` or `passthrough://` rather than `http://` and https://`

pkg/sync/syncWorker.go
richardhuaaa: Yes, first we need to have subscriptions that support a cursor, and then implement tracking of lastSeen (done at the same time as validation of originator payloads) - we have work items for these

pkg/sync/syncWorker.go
```diff
@@ -0,0 +1,205 @@
...
+				"Stream closed with error: %v",
+				err)
+		}
+		for _, env := range envs.Envelopes {
+			s.insertEnvelope(env)
+		}
+	}
+}
+
+func (s *syncWorker) insertEnvelope(env *message_api.OriginatorEnvelope) {
```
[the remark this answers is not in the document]
richardhuaaa: There are two types of failure:
1. An internal error on the server, for example the DB is inaccessible - we need to retry those indefinitely.
2. The data we received from another node is invalid, for example marshalled or signed incorrectly, or in the wrong sequence. For those we will file a `MisbehaviorReport` and continue.

I'll handle both as part of the task `Validation of originator payloads (sid, timestamp, signature), with misbehavior reports`

pkg/sync/syncWorker.go
richardhuaaa: How about I change it to `Warn`? The reason for splitting off the `syncService` and `syncWorker` is so that we can have a single worker per cluster, so it's unexpected to see this in normal operation

pkg/sync/syncWorker.go
richardhuaaa: Okay - I've changed the `syncWorker`, `subscribeWorker` and `publishWorker` to all do this

pkg/sync/syncWorker.go
richardhuaaa: Yup, it's logged on L79 above

richardhuaaa: Just realized I'm re-declaring err here, will fix that

pkg/sync/syncWorker.go
```diff
@@ -0,0 +1,205 @@
...
+	err = proto.Unmarshal(env.GetUnsignedOriginatorEnvelope(), unsignedEnvelope)
+	if err != nil {
+		log.Error(
+			"Failed to unmarshal unsigned originator envelope",
+			zap.Error(err),
+		)
+	}
+
+	clientEnvelope := &message_api.ClientEnvelope{}
+	err = proto.Unmarshal(
```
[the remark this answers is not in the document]
richardhuaaa: Whoops, 100% right. Early returning on these, will add proper test cases once I've implemented proper validation

pkg/sync/syncWorker.go
neekolas: Are you sure about that? With the `:=` on this line wouldn't you be defining a new `err` variable scoped to the loop, not re-defining the `err` defined on line 76.

For example, this code will print `nil`:
```golang
func main() {
	var err error
	for i := 0; i < 5; i++ {
		fmt.Println(err)
		err := fmt.Errorf("foo")
	}
}
```

neekolas: 👍

Your comment appeared just after I hit submit

pkg/sync/syncWorker.go
mkysel: We should probably apply the same logic to sync workers as we do to the grpc servers. Either we guarantee that there is only ever 0-1 (by shutting it down first and breaking sync for a few minutes) or we have to assume there are 2+.

pkg/sync/syncWorker.go
mkysel: the current registry is still full of invalid <100 nodes. Luckily that should be caught by the validity check.

neekolas (approved):

pkg/sync/syncWorker.go
richardhuaaa: I think it makes sense to assume that there will be 2 during upgrades - that will log a few warnings here, and then nothing else in normal operation

PR #213, opened 2024-10-09
Create signing method

neekolas:
## tl;dr
The most widely used [JWT library in Go](https://pkg.go.dev/github.com/golang-jwt/jwt/v5#example-Parse-Hmac) doesn't support secp256k1 signatures natively, so this adds the support by registering a new signing method

## AI Generated Summary

Implemented a custom JWT signing method for secp256k1 and added necessary dependencies.

### What changed?

- Added `github.com/golang-jwt/jwt/v5` dependency to `go.mod` and `go.sum`.
- Created a new `SigningMethodSecp256k1` struct in `pkg/authn/signingMethod.go` that implements a custom JWT signing method for secp256k1.
- Implemented `Verify`, `Sign`, and `Alg` methods for the custom signing method.
- Added unit tests in `pkg/authn/signingMethod_test.go` to verify the functionality of the new signing method.

### How to test?

1. Run the newly added unit tests in `pkg/authn/signingMethod_test.go`.
2. Verify that all tests pass, including:
   - Signing and verifying a message
   - Testing with wrong signer
   - Testing with wrong signing string
   - Full JWT flow with custom signing method

### Why make this change?

This change introduces a custom JWT signing method for secp256k1, which is not natively supported by the standard JWT library. This allows for the use of Ethereum-compatible keys in JWT authentication, enhancing the compatibility of the authentication system with blockchain-based identities.

### DIFF
--- a/pkg/authn/signingMethod.go
+++ b/pkg/authn/signingMethod.go
@@ -1,0 +1,88 @@
+package authn
+
+import (
+	"crypto/ecdsa"
+	"errors"
+	"math/big"
+
+	ethcrypto "github.com/ethereum/go-ethereum/crypto"
+	"github.com/golang-jwt/jwt/v5"
+)
+
+const (
+	ALGORITHM               = "ES256K"
+	SIG_LENGTH              = 65
+	R_LENGTH                = 32
+	S_LENGTH                = 32
+	DOMAIN_SEPARATION_LABEL = "jwt"
+)
+
+var (
+	ErrWrongKeyFormat = errors.New("wrong key type")
+	ErrBadSignature   = errors.New("bad signature")
+	ErrVerification   = errors.New("signature verification failed")
+)
+
+/*
+*
+The JWT signing method for secp256k1. Inspired by https://github.com/ureeves/jwt-go-secp256k1/blob/master/secp256k1.go
+but updated to work with the latest version of jwt-go.
+*/
+type SigningMethodSecp256k1 struct{}
+
+func (sm *SigningMethodSecp256k1) Verify(signingString string, sig []byte, key interface{}) error {
+	pub, ok := key.(*ecdsa.PublicKey)
+	if !ok {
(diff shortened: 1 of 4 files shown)


neekolas:
* **#213** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/213?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pkg/authn/signingMethod.go
```diff
@@ -0,0 +1,91 @@
...
+
+	return sig, nil
+}
+
+func (sm *SigningMethodSecp256k1) Alg() string {
+	return ALGORITHM
+}
+
+func hashString(signingString string) []byte {
+	hasher := crypto.SHA256.New()
```
richardhuaaa, line 80: We use `keccak256` in `registrant.go`, which I think is also consistent with ERC191 and personal_sign. Would it make sense to share the same ciphersuite?

richardhuaaa: If we do this, one thing we need to be careful about is having good domain separation, so that signatures from one use-case cannot be applied to the other.

neekolas: That’s a good call. I can change out the hash function.

neekolas: > If we do this, one thing we need to be careful about is having good domain separation, so that signatures from one use-case cannot be applied to the other.

It should be impossible to have a valid payload that is both a JWT and an UnsignedOriginatorEnvelope, but we could add some prefix to both payloads before hashing so we never have to think about it again.

neekolas: I added a domain separation label just to make sure it's truly impossible to re-use a signature (outside of hash collisions)

richardhuaaa (approved):
It would have been nice to share an implementation with `registrant.go`'s `signKeccak256`, but either way is fine!

PR #219, opened 2024-10-12
Add support for batch query

### LINKED ISSUE #218: Support batch queries

richardhuaaa:
Implements batch querying and unit tests, which will also be used in the next PR for catch-up of historical messages when streaming. Also got rid of our use of NullInt32 in queries, which simplifies the go code a little.

For now we let up to 10k topics be specified in these queries - we can do proper benchmarking and SQL analyzing later to see how we should adjust it.

Closes #218

### DIFF
--- a/pkg/db/queries.sql
+++ b/pkg/db/queries.sql
@@ -30,20 +30,18 @@ FROM
 	gateway_envelopes
 	-- Assumption: There is only one cursor per node ID. Caller must verify this
 	LEFT JOIN cursors ON gateway_envelopes.originator_node_id = cursors.cursor_node_id
-WHERE (sqlc.narg('topic')::BYTEA IS NULL
-	OR length(@topic) = 0
-	OR topic = @topic)
-AND (sqlc.narg('originator_node_id')::INT IS NULL
-	OR originator_node_id = @originator_node_id)
-AND (cursor_sequence_id IS NULL
-	OR originator_sequence_id > cursor_sequence_id)
+WHERE (COALESCE(ARRAY_LENGTH(@topics::BYTEA[], 1), 0) = 0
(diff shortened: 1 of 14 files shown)


pkg/db/queries.sql
```diff
@@ -30,20 +30,18 @@ FROM
...
 	-- Assumption: There is only one cursor per node ID. Caller must verify this
 	LEFT JOIN cursors ON gateway_envelopes.originator_node_id = cursors.cursor_node_id
-WHERE (sqlc.narg('topic')::BYTEA IS NULL
-	OR length(@topic) = 0
-	OR topic = @topic)
-AND (sqlc.narg('originator_node_id')::INT IS NULL
-	OR originator_node_id = @originator_node_id)
-AND (cursor_sequence_id IS NULL
-	OR originator_sequence_id > cursor_sequence_id)
+WHERE (COALESCE(ARRAY_LENGTH(@topics::BYTEA[], 1), 0) = 0
```
neekolas, line 33: What if we broke this into two separate queries and had the caller choose the right one? Saves us from putting the topic list in the query twice, and would be a little more readable.

neekolas (approved):

pkg/db/queries.sql
richardhuaaa: I just tried going down that path, but found it increased complexity elsewhere - spreading if/else clauses throughout the go code. We'd have to break into three queries (as there is also a variant where neither topics nor originators are specified), and all three queries have different return values according to sqlc.

I understood the point about not passing the topic list twice to the database though - what I did instead was refactor the query into a postgres function, so that each param is only passed once to the database. Also used local postgres variables to improve readability slightly.

PR #222, opened 2024-10-15
Allow override of DB id

mkysel:
During development it seems to be pretty common to change the nodeId from one to another. We have to nuke the DB to fix this.

Can we simply change the NodeId in the DB without catastrophic consequences?


neekolas (commented):
In the case where the node ID changes, we want to do more than just skip the check. We want all messages in the database to be wiped. Otherwise we can get into a weird situation where you have messages originated from a node ID that doesn't exist, and other peers who are syncing from this node will have inconsistent records. That will make debugging sync issues much harder.

richardhuaaa:
Marking this as draft for now - I think what we discussed was creating separate DB namespaces for different nodeID's!

PR #223, opened 2024-10-15
Onboarding doc

mkysel:
Add onboarding doc.

Also print address with `cli gen-key`

### DIFF
--- a/doc/onboarding.md
+++ b/doc/onboarding.md
@@ -1,0 +1,58 @@
+# How to onboard a new node
+
+**⚠️ Experimental:** warning, this file might be out of date!
+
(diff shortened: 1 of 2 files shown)


doc/onboarding.md
```diff
@@ -0,0 +1,58 @@
...
+  "address": "0x9419db765e6b469edc028ffa72ba2944f2bad169"
+}
+
+```
+
+## Step 2) Share `pub-key` and `address`
+
+TBD.
+
+Only the members of @ephemerahq/backend can register nodes.
```
richardhuaaa, line 39:
```suggestion
Before official testnet launch, only the members of @ephemerahq/backend can register nodes.
```

richardhuaaa (approved):

PR #230, opened 2024-10-16
Fix GRPC TLS

mkysel:
Load TLS certs from host machine if possible.
Determine whether TLS should be used based on HTTPS scheme.
`passthrough` and `dns` are technically not required.

All tests still pass. A local node can now connect to `testnet`. Local nodes can peer.


richardhuaaa (approved):
Wasn't aware we were already using TLS! Great work

PR #234, opened 2024-10-16
Recover signer from payer envelopes

neekolas:
## tl;dr

- Adds method to recover the signer from payer envelopes (https://github.com/xmtp/xmtpd/issues/232)
- Refactors domain separation for jwt's and payer envelopes

### AI Assisted Summary

Introduced domain separation for payer signatures, and added functionality to recover signer addresses from payer envelopes.

### What changed?

- Created a new `constants` package with domain separation labels for JWT and payer signatures.
- Updated the JWT signing method to use the new constant for domain separation.
- Implemented a `RecoverSigner` method for `PayerEnvelope` to extract the signer's address.
- Added utility functions for hashing and signing payer envelopes.
- Expanded test coverage for envelope signer recovery.

### Why make this change?

This change enhances security and consistency by introducing domain separation for signatures. It also adds the ability to recover signer addresses from payer envelopes, which is crucial for verifying the authenticity of messages and implementing payer-based features in the system.

### DIFF
--- a/pkg/authn/signingMethod.go
+++ b/pkg/authn/signingMethod.go
@@ -7,14 +7,14 @@ import (
 
 	ethcrypto "github.com/ethereum/go-ethereum/crypto"
 	"github.com/golang-jwt/jwt/v5"
+	"github.com/xmtp/xmtpd/pkg/utils"
 )
 
 const (
-	ALGORITHM               = "ES256K"
-	SIG_LENGTH              = 65
-	R_LENGTH                = 32
-	S_LENGTH                = 32
-	DOMAIN_SEPARATION_LABEL = "jwt"
+	ALGORITHM  = "ES256K"
+	SIG_LENGTH = 65
+	R_LENGTH   = 32
+	S_LENGTH   = 32
 )
 
 var (
@@ -36,7 +36,7 @@ func (sm *SigningMethodSecp256k1) Verify(signingString string, sig []byte, key i
 		return ErrWrongKeyFormat
 	}
 
-	hashedString := hashStringWithDomainSeparation(signingString)
+	hashedString := utils.HashJWTSignatureInput([]byte(signingString))
 
 	if len(sig) != SIG_LENGTH {
 		return ErrBadSignature
@@ -58,7 +58,7 @@ func (sm *SigningMethodSecp256k1) Sign(signingString string, key interface{}) ([
 		return nil, ErrWrongKeyFormat
 	}
 
-	hashedString := hashStringWithDomainSeparation(signingString)
+	hashedString := utils.HashJWTSignatureInput([]byte(signingString))
 
 	sig, err := ethcrypto.Sign(hashedString, priv)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


neekolas:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/234?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#234** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/234?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#233** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/233?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

pkg/constants/constants.go
```diff
@@ -0,0 +1,6 @@
+package constants
+
+const (
+	JWT_DOMAIN_SEPARATION_LABEL   = "jwt|"
+	PAYER_DOMAIN_SEPARATION_LABEL = "payer|"
```
neekolas, line 5: I'm going to add domain separation for originator envelopes as well in a separate PR

pkg/authn/signingMethod.go
```diff
@@ -7,14 +7,14 @@ import (
...
+	"github.com/xmtp/xmtpd/pkg/utils"
 )
 
 const (
-	ALGORITHM               = "ES256K"
-	SIG_LENGTH              = 65
-	R_LENGTH                = 32
-	S_LENGTH                = 32
-	DOMAIN_SEPARATION_LABEL = "jwt"
+	ALGORITHM  = "ES256K"
```
richardhuaaa, line 14: Kind of fiddly at this point, but I believe the definition of `ES256K` specifies that sha256 is used as the hashing algorithm. I guess there's no JWT `ALGORITHM` that uses keccak256?

pkg/constants/constants.go
richardhuaaa: Wondering if we can make these shorter, seeing as they're on every signature (which is included 2x per envelope)? And would this make sense as 1 or more bytes rather than a string, seeing as it's concatenated to bytes?

richardhuaaa (approved):

pkg/authn/signingMethod.go
neekolas: Ya. I noticed that too.

We could make one up, but I think this is less confusing.

pkg/constants/constants.go
neekolas: The size kinda comes free, since it just gets added before the inputs are hashed and doesn't go into the signature itself.

neekolas: > And would this make sense as 1 or more bytes rather than a string, seeing as it's concatenated to bytes?

Go [doesn't let you use byte arrays as constants](https://blog.boot.dev/golang/golang-constant-maps-slices/). The alternative would be to define a function like `GetPayerDomainSeparationLabel` and have it return the value as a byte array. But I don't think that's any cleaner than this.

neekolas:
### Merge activity

* **Oct 17, 5:18 PM PDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/xmtp/xmtpd/234).

PR #252, opened 2024-10-23
Handle registry changes

mkysel:
There are some drive-by changes to how we log closing events. It was info, which is okay, but you don't usually need to print 30 lines when shutting down a service. So I moved all of them to debug. You need to be actively debugging shutdown issues to need those.

I fixed shutdown. It was pretty broken. The large issue was that we were waiting on server shutdown in various blocking methods before we actually called the context cancel method. Now we first cancel the context and then go into the various shutdown routines. This, amongst other improvements, means that shutdown now doesn't have to timeout and is much faster.

The meat of this changeset is the introduction of registry listeners. The challenge there is that we don't hold a lock on the registry between calls so I had to refactor a few things to be thread safe. I always chose the pattern of establishing a channel/registry/listener first, then getting the newest state, then acting on that state.

Another annoying part is that we can only rebuild the sync client by cancelling the context. Otherwise it will just sit in Recv() and never wake up. This means that it can't be a simple channel and it was a slightly bigger surgery.

I have tested this manually and it behaves as expected. It needs a follow up combo of unit tests. We also probably need to introduce larger node2node tests. The blockchain is very slow, changes take solid 20+ seconds, so its not practical for integration tests. Hopefully we can just use the mock registry.

## Summary by CodeRabbit

- **New Features**
	- Added subscription management improvements in the `syncWorker` for better handling of node subscriptions.

- **Bug Fixes**
	- Enhanced error handling during the shutdown process of the `ReplicationServer`.

- **Refactor**
	- Simplified the `get-all-nodes` command by removing unnecessary parameters.
	- Updated logging levels across various components to improve verbosity and clarity.

- **Tests**
	- Added a skip message to the `TestCreateServer` function indicating that the test is currently broken.


neekolas:
@mkysel I just changed the settings so that the bot will review draft PRs as well. That will let us get feedback from the bot before reviewers get automatically tagged.

pkg/server/server.go
```diff
@@ -183,15 +183,21 @@ func (s *ReplicationServer) WaitForShutdown() {
 }
 
 func (s *ReplicationServer) Shutdown() {
+	// cancel the context first so that all listeners/servers start shutting down
+	s.cancel()
```
neekolas, line 187: We'd want to make sure this doesn't break graceful shutdown of the API, since a lot of the dependencies of the API service will start to shut down once the context is canceled (for example, the metrics server, the node registry and registrant, and the sync service). I believe the DB will still be up because it is created with a parent context of the one being canceled.

pkg/sync/syncWorker.go
```diff
@@ -51,57 +63,184 @@ func startSyncWorker(
...
 				}
 			}
+		})
+}
+
+func (s *syncWorker) subscribeToNodeInner(ctx context.Context, nodeid uint32) {
+
+	notifierCtx, notifierCancel := context.WithCancel(ctx)
+
+	node, err := s.setupNodeRegistration(notifierCancel, nodeid)
```
richardhuaaa, line 150: Is this method idempotent? What happens if this succeeds but `handleNodeConnection` fails, and we call this again on retry? Will we have multiple goroutines listening for changes on the same node?

pkg/sync/syncWorker.go
```diff
@@ -51,57 +63,184 @@ func startSyncWorker(
...
+
+func (s *syncWorker) handleNodeConnection(
+	notifierCtx context.Context,
+	node *registry.Node,
+) {
+	var conn *grpc.ClientConn
+	var stream message_api.ReplicationApi_SubscribeEnvelopesClient
+	var err error
+
+	//mkysel we should eventually implement a better backoff strategy
```
richardhuaaa, line 171: `// TODO(mkysel)` makes it a bit easier to grep for, I have a bunch of these in the code too

pkg/sync/syncWorker.go
```diff
@@ -51,57 +63,184 @@ func startSyncWorker(
...
+		s.ctx,
+		&s.wg,
+		fmt.Sprintf("node-subscribe-%d", nodeid),
+		func(ctx context.Context) {
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				default:
+					s.subscribeToNodeInner(ctx, nodeid)
```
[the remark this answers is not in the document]
mkysel: no. It is to re-establish the connection after the registry changed.

pkg/sync/syncWorker.go
```diff
@@ -51,57 +63,184 @@ func startSyncWorker(
...
+
+	notifierCtx, notifierCancel := context.WithCancel(ctx)
+
+	node, err := s.setupNodeRegistration(notifierCancel, nodeid)
+	if err != nil {
+		return
+	}
+
+	if !node.IsHealthy || !node.IsValidConfig {
+		s.handleUnhealthyNode(notifierCtx)
```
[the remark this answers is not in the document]
mkysel: can you be more specific? This function only returns if the configuration changes, or if the context shuts down.

pkg/sync/syncWorker.go
mkysel: handleNodeConnection has no failure modes. It only ever returns if the context cancels. Which is only if the config changes, or if we shut down.

pkg/server/server.go
mkysel: ok. its last again. I pulled out a few other services that needed explicit shutting down.

pkg/sync/syncWorker.go
```diff
@@ -51,57 +68,185 @@ func startSyncWorker(
...
 				}
 			}
+		})
+}
+
+func (s *syncWorker) subscribeToNodeInner(ctx context.Context, nodeid uint32) {
+
+	notifierCtx, notifierCancel := context.WithCancel(ctx)
+
+	node, err := s.setupNodeRegistration(notifierCancel, nodeid)
```
richardhuaaa, line 156: I'm understanding everything much better now - I wonder if a few small readability changes might help obtuse people like myself, something like the following:

- Pull out a `NodeConfiguration` struct with `node` and `ctx` stored internally - this replaces `node`, `notifierCtx` and `notifierCancel` here
- `setupNodeRegistration()`->`getCurrentNodeConfiguration()` returning a `NodeConfiguration`, and moved out into the `default` clause of the loop in `subscribeToNode()`.
- `subscribeToNodeInner()`->`subscribeToNodeConfiguration()` that accepts a `NodeConfiguration` argument
- Flatten `handleUnhealthyNode()` into `subscribeToNodeConfiguration()` (would just be `<-nodeConfig.ctx.Done()`)
- Set up the stream in `handleNodeConnection()` using `nodeConfig.ctx`

Will leave it up to your judgment!

richardhuaaa (approved):
Great job handling a complex state machine, and smoothing over the issues with tearing down GRPC streams

pkg/sync/syncWorker.go
mkysel: thank you! It's much cleaner that way!

PR #260, opened 2024-10-24
Set GRPC keep-alive

richardhuaaa:
Our sync streams are being terminated every 1 minute, possibly by the load balancer, possibly because the connection is idle. Here we set GRPC keep-alives to keep the connection active. Let's see if this fixes the issue.

Also added a log to log what a node is registered as.

More info:
- https://github.com/grpc/grpc-node/issues/1747
- https://grpc.io/docs/guides/keepalive/
- https://github.com/grpc/grpc-go/blob/master/Documentation/keepalive.md

## Summary by CodeRabbit

- **New Features**
	- Enhanced logging for registrant identification to improve visibility during runtime.
	- Updated gRPC client connection settings with keepalive parameters for better connection management.

- **Bug Fixes**
	- None reported.

- **Documentation**
	- None reported.

### DIFF
--- a/pkg/registry/node.go
+++ b/pkg/registry/node.go
@@ -3,9 +3,11 @@ package registry
 import (
 	"crypto/ecdsa"
 	"fmt"
+	"time"
 
 	"github.com/xmtp/xmtpd/pkg/utils"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/keepalive"
 )
 
 type DialOptionFunc func(node Node) []grpc.DialOption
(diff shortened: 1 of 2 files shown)


pkg/registry/node.go
```diff
@@ -49,6 +51,11 @@ func (node *Node) BuildClient(
 	dialOpts := append([]grpc.DialOption{
 		grpc.WithTransportCredentials(creds),
 		grpc.WithDefaultCallOptions(),
+		grpc.WithKeepaliveParams(keepalive.ClientParameters{
```
neekolas, line 54: In `server.go` we have a `KeepAliveEnforcementPolicy` set with a `MinTime` of 15 seconds. I believe we will start booting clients who do keepalives more frequently than that. 30 secs should be fine

richardhuaaa: Nice catch

mkysel (approved):

neekolas (approved):

rileymiller:
@richardhuaaa did this wind up fixing your issue?

richardhuaaa:
Yes

PR #271, opened 2024-10-29
Terminate self on prolonged chain disconnect

mkysel:
If the chain is unreachable during startup, the node will refuse to come up since it can't determine its nodeid.
If the chain goes away (for example due to a partial network failure) the node can drift too far away from the ordered MLS messages. We can prevent that scenario from happening by shutting ourselves down.

There are generally 3 ways to shut down a node:
- cancel the main context
- signal self
- panic

I deliberately picked `panic` here as it is the most visible of the three. It will print a stack trace which usually indicates that something is not ok. Its much more "visible" than an error message and a clean shut down.

An orchestrator will restart the node but if the disconnect persists, it will not come up.

The default self termination timeout is 5 min during which we will attempt to re-connect every 100ms. A network that does not heal within 5min sounds like a larger outage.

Fixes https://github.com/xmtp/xmtpd/issues/267

## Summary by CodeRabbit

- **New Features**
	- Introduced a new configuration option for maximum allowable disconnect time in log streaming, enhancing connection stability.
  
- **Bug Fixes**
	- Improved error handling related to connection stability during log streaming sessions.

- **Documentation**
	- Updated documentation to reflect the addition of the `MaxChainDisconnectTime` parameter across various configurations and methods.

- **Tests**
	- Modified test configurations to include the new `MaxChainDisconnectTime` parameter, ensuring comprehensive testing of the new functionality.

### DIFF
--- a/pkg/blockchain/rpcLogStreamer.go
+++ b/pkg/blockchain/rpcLogStreamer.go
@@ -115,7 +115,7 @@ func (r *RpcLogStreamer) Start() {
 }
 
 func (r *RpcLogStreamer) watchContract(watcher contractConfig) {
-	fromBlock := int(watcher.fromBlock)
+	fromBlock := watcher.fromBlock
 	logger := r.logger.With(zap.String("contractAddress", watcher.contractAddress.Hex()))
 	startTime := time.Now()
 	defer close(watcher.channel)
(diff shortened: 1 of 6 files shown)


neekolas (approved):
This makes sense to me. I’d much rather fail loudly than silently.

PR #273, opened 2024-11-01
Pre-baked Anvil+XMTPD docker image

mkysel:
The purpose of this image is it's usage in `libxmtp` without having to pull all the requirements.

We now automatically build a `ghcr.io/xmtp/anvil-xmtpd` image that contains 2 endpoints (localhost:5050 and localhost:5051) along with the default test keys.

If we like this pattern, I will make pre-baked images for `xmtpd` too.

Maybe we can release/tag these in sync so that no configuration is required.

## Summary by CodeRabbit

- **New Features**
	- Introduced a GitHub Actions workflow to automate Docker image building and pushing.
	- Added a new Dockerfile for building an image with the Foundry framework and XMTPD Node contracts.
	- Created a shell script to run the Anvil tool in the background.

- **Documentation**
	- Updated metadata labels in the Dockerfile for better maintainability.

### DIFF
--- a/dev/baked/Dockerfile
+++ b/dev/baked/Dockerfile
@@ -24,18 +24,14 @@ RUN dev/docker/anvil-background && \
 
 # ACTUAL IMAGE -------------------------------------------------------
 
-FROM debian:bookworm
+FROM ghcr.io/foundry-rs/foundry
 
 LABEL maintainer="engineering@xmtp.com"
 LABEL source="https://github.com/xmtp/xmtpd"
-LABEL description="XMTP Node Software"
+LABEL description="Foundry with XMTPD Node contracts and registrations"
 
-# color, nocolor, json
-ENV GOLOG_LOG_FMT=nocolor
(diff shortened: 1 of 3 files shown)


dev/baked/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+SHELL ["/bin/bash", "-c"]
+
+RUN apt-get update &&  \
+    apt-get install -y curl git jq && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN curl -L https://foundry.paradigm.xyz | bash && \
+    source ~/.bashrc && \
+    foundryup -v nightly-2044faec64f99a21f0e5f0094458a973612d0712 &&  \
```
neekolas, line 16: Can we pull the version out to a build arg to make it more visible and easier to maintain? Then we can have the version live in one place and use it everywhere

neekolas (approved):

PR #290, opened 2024-11-25
Add option to manually enable indexer

neekolas:
## tl;dr

- Makes the blockchain indexer optional, so that we can only run it on the same instance that runs the sync worker
- This will require an infra change to set the config

## Summary by CodeRabbit

- **New Features**
	- Introduced a new configuration option to enable or disable the indexer, enhancing user control over system resources.
- **Bug Fixes**
	- Implemented conditional logic to prevent unnecessary indexer initialization, improving resource management and reducing potential errors.

### DIFF
--- a/pkg/config/options.go
+++ b/pkg/config/options.go
@@ -27,6 +27,10 @@ type DbOptions struct {
 	WaitForDB              time.Duration `long:"wait-for"                 env:"XMTPD_DB_WAIT_FOR"                 description:"wait for DB on start, up to specified duration"`
 }
 
+type IndexerOptions struct {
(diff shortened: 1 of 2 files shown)


richardhuaaa (approved):

mkysel:
What is the required combo of flags? Does the indexer make sense without the sync service? Can the sync service run without it? If the answer is that both must run together or neither should be run, it should be just one flag.

neekolas:
They can run independently, we are just likely to choose for them to be run together because that's the service with a desired capacity of 1. But we could break them into two separate services if we chose to.

neekolas:
### Merge activity

* **Nov 25, 5:39 PM PST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/xmtp/xmtpd/290).

PR #303, opened 2024-12-04
Custom flag validation

mkysel:
## Summary by CodeRabbit

- **New Features**
	- Introduced command-line options for various functionalities, including key generation, node registration, and health updates.
	- Added validation for server configuration options to enhance error handling and user feedback.

- **Bug Fixes**
	- Improved error handling for option parsing to provide clearer messages for help and parsing errors.

- **Chores**
	- Simplified the structure of command-line options by consolidating related fields, improving organization and reducing complexity.
	- Updated database initialization logic to conditionally create instances based on enabled features.
	- Enhanced server initialization logic for clearer service management.


cmd/replication/main.go
```diff
@@ -57,16 +63,19 @@ func main() {
...
-			options.DB.WriterConnectionString,
-			utils.BuildNamespace(options),
-			options.DB.WaitForDB,
-			options.DB.ReadTimeout,
-		)
-
-		if err != nil {
-			logger.Fatal("initializing database", zap.Error(err))
+		var dbInstance *sql.DB
+		if options.Replication.Enable || options.Sync.Enable || options.Indexer.Enable {
```
richardhuaaa, line 67: I'm wondering if it makes sense to use a singleton pattern for the DB, and just instantiate it wherever it is needed, rather than trying to comprehensively enumerate all the cases here.

Or: Just instantiate the DB if the DB params are specified, and don't instantiate if it's not, and throw an error if a DB is used somewhere when it hasn't been instantiated.

richardhuaaa (approved):

cmd/replication/main.go
mkysel: I will give it some thought tomorrow!

PR #307, opened 2024-12-05
Build CLI Docker Image

mkysel:
Partners might have to run the CLI to export the public key (or generate it in the first place) and provide it to Ephemera for use in the `testnet`.

It is not too convenient to download the repo, install go, just to run the CLI.

Build a docker image instead.

We will eventually need releases here in GH for easy download, and `homebrew` as well.

The important question is. How should this binary be called?

This should help @jhaaaa with docs.


richardhuaaa (approved):
xmtpd-cli sounds good to me!

Issue #302, opened 2024-12-03
Validation of originator payloads (sid, timestamp, signature), with misbehavior reports


richardhuaaa:

neekolas:
TODO @neekolas tackle some of the TODOs from #311

PR #311, opened 2024-12-06
Validate node ID, sequence ID and timestamp on synced payloads

richardhuaaa:
This adds initial validation on synced payloads.

Still to be done:

- Misbehavior reporting system
- Payload-specific validation

#302 

## Summary by CodeRabbit

- **New Features**
	- Enhanced query logic to ensure unique results for originator nodes.
	- Added a method to access `OriginatorNs` value in the `OriginatorEnvelope` struct.
	- Improved stream management and envelope processing in the sync worker.

- **Bug Fixes**
	- Refined error handling for stream connections and envelope validation.

- **Refactor**
	- Restructured stream and envelope handling to improve data encapsulation and processing logic.


richardhuaaa:
* **#311** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/311?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/311?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

pkg/sync/syncWorker.go
```diff
@@ -318,55 +336,62 @@ func (s *syncWorker) listenToStream(
...
+		lastNs = stream.lastEnvelope.OriginatorNs()
+	}
+	if env.OriginatorSequenceID() != lastSequenceID+1 || env.OriginatorNs() < lastNs {
+		// TODO(rich) Submit misbehavior report and continue
+		s.log.Error("Received out of order envelope")
+	}
+
+	if env.OriginatorSequenceID() > lastSequenceID {
+		stream.lastEnvelope = env
+	}
```
mkysel, line 372: do we check this anywhere at stream startup? What if we rebuild the stream after network failure?

pkg/db/queries.sql
```diff
@@ -46,13 +46,15 @@ DELETE FROM staged_originator_envelopes
 WHERE id = @id;
 
 -- name: SelectVectorClock :many
-SELECT
+SELECT DISTINCT ON (originator_node_id)
```
mkysel, line 49: can you help me understand this change?

pkg/sync/syncWorker.go
richardhuaaa: If you look at where the `originatorStream` is constructed, we are fetching the last payload out of the database before stream construction

pkg/db/queries.sql
richardhuaaa: We are now returning the last envelope as well, so using GROUP BY is not going to work. What we want is one row per originator_node_id, not just the largest sequence ID value.

There's a good section on `DISTINCT ON` in https://www.postgresql.org/docs/current/sql-select.html if that helps!

neekolas (commented):
I’m going to suggest we merge this after we turn off auto-deploy to the testnet, given how big the change is.

pkg/sync/syncWorker.go
```diff
@@ -318,55 +336,62 @@
...
-			OriginatorNodeID: int32(unsignedEnvelope.GetOriginatorNodeId()),
-			OriginatorSequenceID: int64(
-				unsignedEnvelope.GetOriginatorSequenceId(),
-			),
-			Topic:              clientEnvelope.GetAad().GetTargetTopic(),
-			OriginatorEnvelope: originatorBytes,
+			OriginatorNodeID:     int32(env.OriginatorNodeID()),
+			OriginatorSequenceID: int64(env.OriginatorSequenceID()),
+			Topic:                env.TargetTopic().Bytes(),
+			OriginatorEnvelope:   originatorBytes,
```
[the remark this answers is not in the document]
richardhuaaa: We want it to be idempotent here (do nothing if a duplicate already exists)

richardhuaaa:
The golang race test seems to be failing on main (unrelated to this PR), can investigate that separately

pkg/sync/syncWorker.go
```diff
@@ -318,55 +335,62 @@
...
+		// TODO(rich) Submit misbehavior report and continue
+		s.log.Error("Received out of order envelope")
+	}
+
+	if env.OriginatorSequenceID() > lastSequenceID {
+		stream.lastEnvelope = env
+	}
+
+	// TODO Validation logic - share code with API service and publish worker
+	// Signatures, topic type, etc
```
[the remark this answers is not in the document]
richardhuaaa: That's why there is a TODO here, it will be addressed in subsequent PR's as part of a stack

pkg/sync/syncWorker.go
```diff
@@ -318,55 +336,62 @@
...
+	s.insertEnvelope(env)
+}
```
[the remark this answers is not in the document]
richardhuaaa: I think this is a style preference. The caller is never going to do anything with these errors, as they shouldn't break the stream. If there's an invalid envelope, we simply want to log an error and move on.

pkg/sync/syncWorker.go
```diff
@@ -299,15 +306,25 @@ func (s *syncWorker) setupStream(
 			err,
 		)
 	}
-	return stream, nil
+	originatorStream := &originatorStream{nodeID: nodeID, stream: stream}
+	for _, row := range result {
+		if uint32(row.OriginatorNodeID) == nodeID {
+			lastEnvelope, err := envUtils.NewOriginatorEnvelopeFromBytes(row.OriginatorEnvelope)
+			if err != nil {
```
neekolas, line 313: Are there cases where we would actually want to continue here? I worry about an invalid `ClientEnvelope` blocking replication forever because it errors.

Maybe it's the originator's fault for not validating properly before putting it in the DB and we _should_ stop?

Just want to make sure we've considered all the failure modes.

richardhuaaa: Yes, it's a tricky call. I think it's similar to the general programming question of whether we should assert things that should always be true. I'm a bit more in the camp of a simpler programming model where things fail if our assumptions are violated, and we fix them - there are also other places in the codebase that rely on the DB deserialization working (e.g. when a query is made) and it would be preferable not to mask them in our responses to clients.

> Maybe it's the originator's fault for not validating properly before putting it in the DB and we should stop? Would only be an issue if the validation rules changed.

Yes, I think the assumption is that the originator should have serialized it properly before writing to the DB. Note we're not doing any validation on the envelope here - we're just checking that it can be deserialized. So it shouldn't be a problem if the validation rules changed, but it could happen if we make a breaking proto change without preparing it properly.

neekolas: Maybe the safest thing is to leave it as-is and if we have nodes that get jammed up we know we have a problem to fix.

> Note we're not doing any validation on the envelope here - we're just checking that it can be deserialized

FWIW, we do a little bit of validation on the `ClientEnvelope` to make sure the topic matches the payload type. But that all should be checked before it ever goes in the DB.

neekolas (approved):

PR #327, opened 2024-12-16
Print error on failed blockchain publish

mkysel:
I would argue that we should not leak internal errors to the client, but it is actually easier to debug for now.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error logging for blockchain publishing, enhancing visibility of encountered issues.


richardhuaaa (approved):
I think we should add some middleware that automatically logs failed status codes

neekolas (approved):

PR #332, opened 2024-12-18
Introduce interceptor to log client errors

### LINKED ISSUE #329: Add middleware to log client errors
I think we should add some middleware that automatically logs failed status codes

_Originally posted by @richardhuaaa in https://github.com/xmtp/xmtpd/pull/327#pullrequestreview-2507527252_

mkysel:
Here is an example of the logger:
```
2024-12-18T16:17:30.226Z        ERROR   replication     Client Unary RPC Error  {"method": "/xmtp.xmtpv4.payer_api.PayerApi/PublishClientEnvelopes", "duration": 10888708, "code": "Internal", "message": "error publishing group message: rpc error: code = Internal desc = error publishing identity update: timed out"}
```

Fixes #329 

## Summary by CodeRabbit

- **New Features**
	- Introduced a logging interceptor for gRPC services to enhance error logging during RPC calls.
- **Bug Fixes**
	- Improved error handling and logging capabilities for unary and stream RPC calls.
- **Tests**
	- Added unit tests for the logging interceptor to ensure correct logging behavior during error scenarios.

### DIFF
--- a/pkg/interceptors/client/logging.go
+++ b/pkg/interceptors/client/logging.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"fmt"
 	"go.uber.org/zap"
 	"time"
 
@@ -15,14 +16,14 @@ type LoggingInterceptor struct {
 }
 
 // NewLoggingInterceptor creates a new instance of LoggingInterceptor.
-func NewLoggingInterceptor(logger *zap.Logger) *LoggingInterceptor {
+func NewLoggingInterceptor(logger *zap.Logger) (*LoggingInterceptor, error) {
 	if logger == nil {
-		panic("logger is required")
+		return nil, fmt.Errorf("logger is required")
 	}
 
 	return &LoggingInterceptor{
 		logger: logger,
-	}
+	}, nil
 }
 
 // Unary intercepts unary RPC calls to log errors.
--- a/pkg/interceptors/client/logging.go
+++ b/pkg/interceptors/client/logging.go
@@ -1,0 +1,80 @@
+package client
+
+import (
+	"context"
+	"go.uber.org/zap"
+	"time"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/status"
+)
+
+// LoggingInterceptor logs errors for unary and stream RPCs.
+type LoggingInterceptor struct {
+	logger *zap.Logger
+}
+
+// NewLoggingInterceptor creates a new instance of LoggingInterceptor.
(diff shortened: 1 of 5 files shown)


pkg/interceptors/client/logging.go
```diff
@@ -0,0 +1,81 @@
+package client
```
neekolas, line 1: I would think this is a server interceptor, since it only gets attached to server instances.

mkysel: I can move it!

pkg/interceptors/client/logging.go
```diff
@@ -0,0 +1,81 @@
...
+		req interface{},
+		info *grpc.UnaryServerInfo,
+		handler grpc.UnaryHandler,
+	) (interface{}, error) {
+		start := time.Now()
+		resp, err := handler(ctx, req) // Call the actual RPC handler
+		duration := time.Since(start)
+
+		if err != nil {
+			st, _ := status.FromError(err)
```
neekolas: Are we 100% sure the error will always be parse-able by `status.FromError`? I know GRPC converts any error eventually to a GRPC unknown error, but I'm less sure it happens _before_ the interceptors run.

If we got another error type the `st.Code()` call below would panic, since `st` would be nil.

mkysel: Yes. It never returns `nil`
```
func FromError(err error) (s *Status, ok bool) {
	if err == nil {
		return nil, true
	}
	type grpcstatus interface{ GRPCStatus() *Status }
	if gs, ok := err.(grpcstatus); ok {
		grpcStatus := gs.GRPCStatus()
		if grpcStatus == nil {
			// Error has status nil, which maps to codes.OK. There
			// is no sensible behavior for this, so we turn it into
			// an error with codes.Unknown and discard the existing
			// status.
			return New(codes.Unknown, err.Error()), false
		}
		return grpcStatus, true
	}
	var gs grpcstatus
	if errors.As(err, &gs) {
		grpcStatus := gs.GRPCStatus()
		if grpcStatus == nil {
			// Error wraps an error that has status nil, which maps
			// to codes.OK.  There is no sensible behavior for this,
			// so we turn it into an error with codes.Unknown and
			// discard the existing status.
			return New(codes.Unknown, err.Error()), false
		}
		p := grpcStatus.Proto()
		p.Message = err.Error()
		return status.FromProto(p), true
	}
	return New(codes.Unknown, err.Error()), false
}
```

neekolas: Perfect. The `FromError` will always coerce into the right type.

neekolas (approved):

PR #349, opened 2024-12-20
Incoming interceptor

mkysel:
This interceptor will print the incoming connection (and DNS) if it can to the log.

Among other things, this is important for detecting that peers are connected.

## Summary by CodeRabbit

- **New Features**
  - Introduced an `IncomingInterceptor` to log incoming requests in a gRPC server context.
  - Enhanced gRPC server with new logging capabilities for unary and stream calls.

- **Bug Fixes**
  - Improved error handling during the initialization of the incoming interceptor.

- **Tests**
  - Added unit tests for the `IncomingInterceptor`, validating logging functionality for unary and stream requests.

### DIFF
--- a/pkg/interceptors/server/incoming.go
+++ b/pkg/interceptors/server/incoming.go
@@ -1,0 +1,75 @@
+package server
+
+import (
+	"context"
+	"fmt"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/peer"
+	"net"
+)
+
+type IncomingInterceptor struct {
+	logger *zap.Logger
+}
+
+func NewIncomingInterceptor(logger *zap.Logger) (*IncomingInterceptor, error) {
+	if logger == nil {
+		return nil, fmt.Errorf("logger is required")
+	}
+
+	return &IncomingInterceptor{
+		logger: logger,
+	}, nil
+}
+
+func (i *IncomingInterceptor) logIncomingAddressIfAvailable(ctx context.Context) {
+	if i.logger.Core().Enabled(zap.DebugLevel) {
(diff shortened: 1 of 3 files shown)


pkg/interceptors/server/incoming.go
```diff
@@ -0,0 +1,75 @@
...
+			if err == nil {
+				dnsName, err = net.LookupAddr(host)
+				if err != nil || len(dnsName) == 0 {
+					dnsName = []string{"Unknown"}
+				}
+			} else {
+				dnsName = []string{"Unknown"}
+			}
+			i.logger.Debug(
+				fmt.Sprintf("Incoming request from %s (DNS: %s)", clientAddr, dnsName[0]),
```
richardhuaaa, line 42: I understand the desire to log incoming node addresses, but could we avoid logging client addresses?

The best solution of course is for clients to anonymize themselves by routing through an intermediary, and/or the protocol itself doing this. But it would be best if the implementation also did its best to minimize unnecessary metadata collection.

neekolas: Agree the privacy cost of this might not be worth the improved debuggability for client requests.

Client IP addresses are sensitive, and logging them makes every node's log store an attack vector which can be harvested to de-anonymize users.

mkysel: we need something to debug node to node connectivity.

What is your proposed alternative?

neekolas: We could get the node's ID from the auth token and log it, which only applies to node <> node but not client <> node.

mkysel: sounds like a plan

PR #353, opened 2024-12-23
chore: pin abigen version

fbac:
Pin the abigen version, so generated golang code is deterministically generated. 
Different versions could introduce changes, and we should avoid pushing generated code with different versions.

## Summary by CodeRabbit

- **New Features**
	- Introduced version pinning for the `abigen` tool to ensure consistent installation.
	- Added a version check for `abigen` to verify installation integrity.

- **Bug Fixes**
	- Removed the installation command for the `abigen` tool from the script, streamlining the installation process.


mkysel (approved):

neekolas:
The abigen version is set in the go.mod file. This shouldn't be necessary.

fbac:
> The abigen version is set in the go.mod file. This shouldn't be necessary.

As far as I know `go install` works independently and doesn't use the `go.mod` versioning. Am I missing something?

Issue #322, opened 2024-12-16
feat(indexer): handle reorganizations


fbac:
The indexer has to be able of detecting blockchain reorganizations.

In order to provide reorg detection capabilities, the following changes are required:
- the database model has to be expanded and include blockHash, blockNumber and timestamps.
- the indexer has to be able to detect and handle reorganizations.
- the broadcast network has to sync new appended messages created during the reorg handling.

PR #411, opened 2025-01-15
feat: detect chain reorgs

fbac:
Closes [#322](https://github.com/xmtp/xmtpd/issues/322)

## Release Notes

- **New Features**
	- Added support for blockchain reorganization event handling.
	- Enhanced blockchain message tracking with canonicality tracking.
	- Improved blockchain indexing with more robust event management.
	- Introduced new methods for blockchain interactions in mock interfaces.

- **Database Changes**
	- Introduced new `blockchain_messages` table to track block-level metadata.
	- Added support for tracking canonical and non-canonical blockchain events.

- **Performance Improvements**
	- Optimized blockchain event streaming and indexing.
	- Added more granular blockchain state tracking mechanisms.

- **Testing**
	- Expanded mock interfaces for blockchain and indexing components.
	- Added comprehensive test coverage for new reorganization handling logic.

- **Infrastructure**
	- Updated various code generation tool versions.
	- Enhanced blockchain client interface capabilities.
	- Updated SQLC version for improved type handling.


pkg/migrations/00004_update-gateway-envelopes.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE gateway_envelopes
```
mkysel, line 1: we also need a `down` SQL. Basically a revert.

pkg/interceptors/server/auth_test.go
```diff
@@ -36,7 +36,7 @@ func TestUnaryInterceptor(t *testing.T) {
 				return metadata.NewIncomingContext(context.Background(), md)
 			},
 			setupVerifier: func() {
-				mockVerifier.EXPECT().Verify("valid_token").Return(nil)
+				mockVerifier.EXPECT().Verify("valid_token").Return(uint32(0), nil)
```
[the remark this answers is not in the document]
fbac: This is not relevant to the features introduced in this PR.

pkg/migrations/00004_update-gateway-envelopes.sql
fbac: Introduced with the new sql files.

pkg/blockchain/rpcLogStreamer.go
```diff
@@ -48,13 +48,21 @@
...
+			fromBlock,
+			contractAddress,
+			topics,
+			eventChannel,
+			reorgChannel,
+			maxDisconnectTime,
+		},
 	)
-	return eventChannel
+	return eventChannel, reorgChannel
```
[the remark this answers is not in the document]
fbac: ListenForContract is used from indexer.go, there the channel is used as write only. watchContract only has to read from it.

pkg/indexer/indexer_test.go
```diff
@@ -10,59 +10,118 @@ import (
...
 func TestIndexLogsSuccess(t *testing.T) {
 	channel := make(chan types.Log, 10)
-	defer close(channel)
+	reorgChannel := make(chan uint64, 1)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer func() {
+		cancel()
+		close(channel)
+		close(reorgChannel)
+	}()
```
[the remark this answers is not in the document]
fbac: This will be added in a future PR.

pkg/indexer/indexer.go
```diff
@@ -195,20 +215,61 @@ The only non-retriable errors should be things like malformed events or failed v
...
-	var err storer.LogStorageError
+	var errStorage storer.LogStorageError
+
 	// We don't need to listen for the ctx.Done() here, since the eventChannel will be closed when the parent context is canceled
 	for event := range eventChannel {
+		storedBlockNumber, storedBlockHash := blockTracker.GetLatestBlock()
+
+		// TODO: Calculate the blocks safe distance in the L3 or risk tolerance we assume
+		if event.BlockNumber > storedBlockNumber &&
+			event.BlockNumber-storedBlockNumber < safeBlockDistance {
```
neekolas, line 234: We need to balance timeliness with safety here. A lot of these blockchain events are the kind of thing a human is going to be waiting for on their phone. I think that means we need to be optimistic and rely on preconfirmations, and then fix later in the rare emergency case.

pkg/indexer/storer/identityUpdate.go
```diff
@@ -48,7 +50,11 @@ func NewIdentityUpdateStorer(
 }
 
 // Validate and store an identity update log event
-func (s *IdentityUpdateStorer) StoreLog(ctx context.Context, event types.Log) LogStorageError {
+func (s *IdentityUpdateStorer) StoreLog(
+	ctx context.Context,
+	event types.Log,
+	appendLog bool,
```
neekolas, line 56: Maybe we can add a comment about when you would and wouldn't want to use this flag.

pkg/db/queries/queries.sql.go
```diff
@@ -150,6 +173,10 @@ type InsertGatewayEnvelopeParams struct {
 	OriginatorSequenceID int64
 	Topic                []byte
 	OriginatorEnvelope   []byte
+	BlockNumber          sql.NullInt64
+	BlockHash            []byte
```
neekolas, line 177: Just thinking out loud, but what if we did this relationally (like with a table called `block_versions`) and joined to that rather than denormalizing?

Would use less storage and still be easy to find impacted messages in a re-org.

Not a suggestion, more of a question.

mkysel: I made a similar proposal in slack :)

fbac: I like that approach, too. My biggest concern is that for each group/identity message we'd be inserting in db twice and there would be additional storage overhead.

> Would use less storage
I'm not sure I follow, could you expand this?

neekolas: Just that we only have to store the block hash and number once per block, and store a foreign key on each message in the block. The block number and foreign key are the same size so they cancel out, but the block hash is pure savings and bigger.

I'm assuming most blocks have > 1 message in them.

pkg/indexer/indexer.go
fbac: I modified the indexer to be more performant by skipping certain IO tasks when they're not needed:
- Only `blockTracker.GetLatestBlock` when really needed.
- Check for reorgs only after a certain period has passed. The `safeBlockDistance` has been removed in favor of this. Also, I'm not comfortable leaving this open to configurations so I dropped the option to do it.
- When a reorg is detected the indexer will drop the messages that are in the eventChannel, until it starts receiving from the first reorganized block.

pkg/db/queries/queries.sql.go
fbac: SQL schemas updated. It still needs some fine tuning, and better thinking on the queries, but captures the idea.

pkg/indexer/reorgHandler.go
```diff
@@ -0,0 +1,156 @@
...
+		if err != nil && !errors.Is(err, sql.ErrNoRows) {
+			return 0, nil, fmt.Errorf("failed to get stored blocks: %w", err)
+		}
+
+		if len(storedBlocks) == 0 || errors.Is(err, sql.ErrNoRows) {
+			if startBlock == 0 {
+				return 0, nil, ErrNoBlocksFound
+			}
+
+			startBlock, endBlock = blockRange(startBlock)
```
neekolas, line 66: I'm not sure, but is it not possible to get into an infinite loop if the returned `startBlock` is 0, but there are blocks in the `storedBlocks` results for that range?

I don't know how it would happen in practice in a blockchain, but we might want to have an explicit exit if `startBlock` is 0 and we don't have the reorg block inside the range.

pkg/indexer/reorgHandler.go
```diff
@@ -0,0 +1,156 @@
...
+				EndBlock:   endBlock,
+			},
+		)
+		if err != nil && !errors.Is(err, sql.ErrNoRows) {
+			return 0, nil, fmt.Errorf("failed to get stored blocks: %w", err)
+		}
+
+		if len(storedBlocks) == 0 || errors.Is(err, sql.ErrNoRows) {
+			if startBlock == 0 {
+				return 0, nil, ErrNoBlocksFound
```
neekolas, line 63: I worry we are overloading the `ErrNoBlocksFound` error. It seems to mean slightly different things in different places (no blocks in the DB, reorg happened earlier in the chain, empty range)

neekolas (approved):

pkg/indexer/reorgHandler.go
fbac: In every path we check for `startBlock == 0` and return, I don't see how we could get into an infinite loop.

pkg/indexer/reorgHandler.go
fbac: The idea around it is using it when the range starts at 0 and we still didn't find the reorg point, which means the reorg happened since block 0... very unlikely.
Removing it from `searchInRange`, it's true that it doesn't make sense there as we'll be always passing ranges at least length == 1.

pkg/db/queries.sql
```diff
@@ -122,3 +122,35 @@ FROM
...
+	block_number ASC,
+	block_hash;
+
+-- name: UpdateBlocksCanonicalityInRange :exec
+UPDATE
+	blockchain_messages
+SET
+	is_canonical = FALSE
+WHERE
+	block_number >= @reorg_block_number;
```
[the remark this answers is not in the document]
mkysel: @fbac is bunny has the right intuition here. Are we holding a lock somewhere to prevent the insertion of new canonical blocks?

pkg/indexer/storer/groupMessage.go
```diff
@@ -13,6 +13,11 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+const (
+	// We may not want to hardcode this to 0 and have an originator ID for each smart contract?
+	GROUP_MESSAGE_ORIGINATOR_ID = 0
```
mkysel, line 18: I thought it was supposted to be `>=1 ID < 100`...

pkg/indexer/storer/identityUpdate.go
```diff
@@ -177,6 +180,17 @@ func (s *IdentityUpdateStorer) StoreLog(ctx context.Context, event types.Log) Lo
 				return NewLogStorageError(err, true)
 			}
 
+			if err = querier.InsertBlockchainMessage(ctx, queries.InsertBlockchainMessageParams{
```
mkysel, line 183: same atomicity comment here

pkg/migrations/00004_add_blockchain_columns.up.sql
```diff
@@ -0,0 +1,12 @@
...
+	block_number BIGINT NOT NULL,
+	block_hash BYTEA NOT NULL,
+	originator_node_id INT NOT NULL,
+	originator_sequence_id BIGINT NOT NULL,
+	is_canonical BOOLEAN NOT NULL DEFAULT TRUE,
+	PRIMARY KEY (block_number, block_hash, originator_node_id, originator_sequence_id),
+	FOREIGN KEY (originator_node_id, originator_sequence_id) REFERENCES gateway_envelopes(originator_node_id, originator_sequence_id)
+);
+
+CREATE INDEX idx_blockchain_messages_block_canonical ON blockchain_messages(block_number, is_canonical);
```
mkysel, line 11: the `GetBlocksInRange` can't actually use this index because it is missing parts of the query. You need a composite index such as:
```
CREATE INDEX idx_blockchain_messages_block_hash_canonical 
ON blockchain_messages(block_number, is_canonical, block_hash);
```

mkysel: maybe the index without the hash is sufficient and the table scans won't be too bad. Hard to predict without seeing a query plan.

pkg/indexer/reorgHandler.go
```diff
@@ -0,0 +1,152 @@
...
+	client  blockchain.ChainClient
+	queries *queries.Queries
+}
+
+var (
+	ErrNoBlocksFound = errors.New("no blocks found")
+	ErrGetBlock      = errors.New("failed to get block")
+)
+
+// TODO: Make this configurable?
```
mkysel, line 30: Rich always says that TODO should have your name in them :) So we know who thought that there is some work left. You will see `TODO(rich)` and `TODO(mkysel)` through the codebase. I don't have strong opinions on this.

pkg/indexer/indexer.go
```diff
@@ -195,20 +220,120 @@ The only non-retriable errors should be things like malformed events or failed v
...
+			logger.Info("starting processing reorg",
+				zap.Uint64("eventBlockNumber", event.BlockNumber),
+				zap.Uint64("reorgBlockNumber", reorgBeginsAt))
+
+			// When all future events have been discarded, it means we've reached the reorg point
+			storedBlockNumber, storedBlockHash = blockTracker.GetLatestBlock()
+			lastBlockSeen = event.BlockNumber
+			reorgDetectedAt = 0
+			reorgInProgress = true
+		}
```
mkysel, line 264: how does this work in HA? There are multiple indexer processes running. Are they somehow guaranteed to see the same reorg?

mkysel:
Once the reorg handler marks the secondary table as `not canonical` what adjusts the gateway envelopes? All clients interface with the gateway table. Do we delete them? Update them? Stop serving them to clients?

pkg/indexer/storer/groupMessage.go
fbac: To be fair, I just moved the group message origId to its own const, but it was always hardcoded 0. Let me extract an issue out of this.

pkg/db/queries.sql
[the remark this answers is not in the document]
fbac: @mkysel not really, it seems like a good idea holding a lock while setting canonical false locks, before resuming the canonical block pulling.

Issue #358, opened 2025-01-02
Read your own writes consistency


neekolas:

mkysel:
Split into sub-issues.

PR #415, opened 2025-01-15
Read your own group writes

mkysel:
This does not solve all possible cases of read-your-own-writes, but it makes things a little bit easier.

This guarantees that when a node confirms a payer write, future reads will return the data.

This does not guarantee:
- the same for blockchain writes
- a different node will have the data

I added the debug log so we can see how long the delay is. In general, it is usually one busy-wait cycle. But could be more on busy systems. It is rarely 0 cycles.

Relates to #358 

## Summary by CodeRabbit

- **New Features**
	- Added tracking of last processed envelope ID in publish worker
	- Implemented wait mechanism for gateway publish completion
	- Enhanced server test infrastructure with dynamic port allocation

- **Tests**
	- Added new test to verify server's ability to read its own writes immediately after publishing

- **Chores**
	- Improved test configuration by dynamically assigning ports and populating mock configurations

### DIFF
--- a/pkg/api/message/service.go
+++ b/pkg/api/message/service.go
@@ -481,7 +481,7 @@ func (s *Service) waitForGatewayPublish(
 				zap.Int64("envelope_id", stagedEnv.ID),
 				zap.Int64("wait_time", time.Since(startTime).Milliseconds()),
 			)
-			break
+			return
 		}
 		time.Sleep(10 * time.Millisecond)
 	}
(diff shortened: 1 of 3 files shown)


pkg/api/message/service.go
```diff
@@ -452,3 +454,19 @@ func (s *Service) validateClientInfo(clientEnv *envelopes.ClientEnvelope) error
 
 	return nil
 }
+
+func (s *Service) waitForGatewayPublish(stagedEnv queries.StagedOriginatorEnvelope) {
```
neekolas, line 458: We should at least pass the context here so we can return early if the context is canceled. Maybe also add a deadline to the context so this doesn't wait forever if something goes wrong?

neekolas (approved):

Issue #418, opened 2025-01-16
Read-your-own-commits client guarantee


mkysel:
I am splitting the RYOW work into multiple steps.

We already have #415.

This issue represents a similar same node guarantee but for commits that route via the blockchain.

A client, that writes a commit to the payer, should be able to read that commit from a pre-defined node once the payer confirms the write.

To achieve this, we can always use a nodeId (other than 1) that won't necessarily be the originator, but will be the "default" node that the client is communicating with.

This change should guarantee that that node will have indexed the blockchain write.

mkysel:
Spitballing ideas here:
- the payer can read from the target node until the node has processed it
- we don't actually make this guarantee and the client has to retry based on internal timeout rules

mkysel:
My best idea right now is to do the following:
- expose a new endpoint on the API that shows the current newest "cursor" for each nodeID+blockchain
- the payer service keeps an open stream from the node to learn of the newest writes
- use a channel to block the client write until the node has read from the chain

PR #459, opened 2025-02-04
Stable hashing for node selection in the payer

mkysel:
Stable hashing in the payer to pick the correct node to send the write to. Hash is calculated based on the largest known ID and the topic.

This also pulls the changes to protos to stop using the originator_id and use is_commit instead.

Right now the client does not implement is_commit correctly. And hard codes the nodeID to 100 in all cases.

In the case of off-chain messages, this still guarantees that the message was written on that particular node.
In case of on-chain commits, the guarantee introduced in #418 somewhat holds. Since the client is no longer selecting a node, the payer will automatically detect which node should be seen as the preferred node and will block until that node has read the chain.

In neither case can we guarantee that any other node has read any write.

The stable hashing algorithm is pretty naive and might eventually need some enhancements. There are two issues that I see with it as is (which is the standard implementation):
- if there are node ID gaps, the next nodeID gets double or more of the traffic
- if we issue a new node ID it might rehash

The payer does not yet have any retry logic, it does not check the liveness of the nodes, etc. All of that has to happen in a future PR.

Relates to:
- https://github.com/xmtp/XIPs/pull/85
- https://github.com/xmtp/proto/pull/242

Fixes https://github.com/xmtp/libxmtp/issues/1455

## Summary by CodeRabbit

- **New Features**
  - Introduced an enhanced routing mechanism that leverages stable hashing for more optimal node selection and balanced message distribution.

- **Refactor**
  - Standardized how routing data is assigned, ensuring consistent handling of message originators.
  - Refined criteria for routing validation and blockchain publishing, improving overall reliability.

- **Tests**
  - Expanded test coverage to verify stable node selection, robust message routing, and balanced topic assignments.
  - Enhanced test setup to utilize a mock registry for improved clarity and control over interactions.


pkg/api/payer/nodeSelector.go
```diff
@@ -0,0 +1,61 @@
...
+	if len(nodes) == 0 {
+		return 0, errors.New("no available nodes")
+	}
+
+	// Sort nodes to ensure stability
+	sort.Slice(nodes, func(i, j int) bool { return nodes[i].NodeID < nodes[j].NodeID })
+
+	topicHash := HashKey(topic)
+
+	// we do not want to hash the entire available hash space, just consider up to the largest NodeID issued
```
neekolas, line 48: I want to make sure my intuitions are correct here.

Let's say we have nodes 100, 200, and 300.

If we add node 400, what % of the topics will shift to point to a different node? I think by including the highest node ID in the `topicHash` we might defeat the purpose of consistent hashing, which is consistency of selection.

Am I misunderstanding this?

neekolas: Typically when I think of consistent hashing I think of having the hash be `% 360` (or some other constant) and distributing the node IDs equally between 0 and that constant.

mkysel: yes. I point out these limitations in the PR description. I am happy to brainstorm on how to get this right.

neekolas: Just spitballing

We define some constant for the size of the space. Let's say 360.

Then we split the nodes evenly over that space.

```python
spacing = 360 / num_nodes
node_locations = [spacing * index for index, node in enumerate(nodes)] // assumes nodes are sorted by ID
```

Then we can just `hash(topic) % 360` for each topic and put it somewhere in the space. As we add nodes the locations would move but only slightly.

mkysel: the new commit does that. Just in a 32bit space.

pkg/api/payer/nodeSelector_test.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+	}
+	t.Logf("Target Buckets: %v", targetBuckets)
+
+	// Compute expected balance
+	expectedPerNode := totalRequests / uint32(len(nodes))
+	tolerance := float64(expectedPerNode) * 0.05 // 10% tolerance
+	t.Logf("Target Tolerance: %v", tolerance)
+
+	// Verify that each bucket is within ±10% of expected distribution
```
neekolas, line 222: Great test

neekolas (approved):

PR #473, opened 2025-02-10
make indexer handle null block hashes

fbac:
Before the future release 0.2.0 the block_hash is not stored in database. In future releases we're using the block_hash to check for reorgs.

There's a rare corner case where the indexer tries to retrieve a block hash that happens to be null, in that case we should skip it and check for reorgs later in the chain.

In the case when a reorg happens later in the chain, and the reorg handler algorithm determines the reorg happened in a range we don't have any data about, then the indexer should simply start indexing from 0, as it's the safest option.

### DIFF
--- a/pkg/indexer/indexer.go
+++ b/pkg/indexer/indexer.go
@@ -300,7 +300,8 @@ func indexLogs(
 				zap.Uint64("blockNumber", reorgCheckAt),
 			)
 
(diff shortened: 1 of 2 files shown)


mkysel:
LGTM. Do you wanna absorb https://github.com/xmtp/xmtpd/pull/471 into this and merge both together?

mkysel (approved):

Issue #463, opened 2025-02-05
Error: Invalid Target Originator


mkysel:
Right now the client hardcodes originatorID to 100, but that code needs to go away.

If I made an edit to the client such as
```
diff --git a/xmtp_proto/src/convert.rs b/xmtp_proto/src/convert.rs
index 4c003af3..0fe826d0 100644
--- a/xmtp_proto/src/convert.rs
+++ b/xmtp_proto/src/convert.rs
@@ -105,8 +105,7 @@ impl AuthenticatedData {
     #[allow(deprecated)]
     pub fn with_topic(topic: Vec<u8>) -> AuthenticatedData {
         AuthenticatedData {
-            //TODO(mkysel) originator is hardcoded for now, but will have to become configurable
-            target_originator: Some(100),
+            target_originator: None,
             target_topic: topic,
             depends_on: None,
             is_commit: false,
```
the payer will fail:

```
2025-02-05T13:29:49.053-0500	ERROR	replication	error publishing payer envelopes	{"error": "rpc error: code = Internal desc = error publishing payer envelopes: rpc error: code = InvalidArgument desc = invalid target originator"}
2025-02-05T13:29:49.053-0500	ERROR	replication	Client Unary RPC Error	{"method": "/xmtp.xmtpv4.payer_api.PayerApi/PublishClientEnvelopes", "duration": 818125, "code": "Internal", "message": "error publishing payer envelopes"}
```

What I suspect is happening is that the payload goes to an unavailable node and the payer doesn't take unavailable nodes into account.

Since the payer is doing the hashing, it needs to be able to rehash if a node is unavailable.

PR #477, opened 2025-02-11
Validate dependsOn Field

mkysel:
## Summary by CodeRabbit

- **New Features**
  - Introduced an enhanced state tracking mechanism that ensures improved validation and consistency during message processing and replication.
  - Added a cursor management feature to the replication and metadata services.

- **Refactor**
  - Streamlined service initialization and dependency management for more reliable performance and graceful resource cleanup.

- **Tests**
  - Enhanced the test suite with new scenarios to validate error handling related to cursor dependencies in the publishing process.

### DIFF
--- a/pkg/api/message/service.go
+++ b/pkg/api/message/service.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"github.com/xmtp/xmtpd/pkg/api/metadata"
 	"time"
 
 	"github.com/xmtp/xmtpd/pkg/db"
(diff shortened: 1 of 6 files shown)


pkg/api/message/service.go
```diff
@@ -449,7 +453,25 @@ func (s *Service) validateClientInfo(clientEnv *envelopes.ClientEnvelope) error
 		return status.Errorf(codes.InvalidArgument, "topic does not match payload")
 	}
 
-	// TODO(rich): Verify all originators have synced past `last_seen`
+	if aad.GetDependsOn() != nil {
```
fbac, line 456: LGTM.
Should we remove the `aad.GetTargetOriginator() != s.registrant.NodeID()` check on this PR?

fbac (approved):

pkg/api/message/service.go
mkysel: Yeah. That check will have to go. I need to introduce the new field set by the payer. I suspect we can still validate the originator if it exists for backwards compatibility?

neekolas: Burn the boats. We're too early to worry about backward compatibility for things like validations that don't require coordination between nodes..

mkysel: It's tracked here https://github.com/xmtp/xmtpd/issues/463

PR #481, opened 2025-02-12
tie xmtpd to specific proto versions

fbac:
Currently we generate the protos based on the main branch. 
Instead of doing this, and to help with reproducible deterministic build each xmtpd tag should point to the latest supported tag in xmtp/proto.

Suggestion:
- While developing in main, point to ref=latest_supported_commit
- When publishing a release, point to tag=vLatest_supported_proto_tag

## Summary by CodeRabbit

- **Chore**
  - Adjusted the process for generating critical application definitions by using a fixed version reference, ensuring a more stable and consistent underlying build process.


mkysel:
We don't autogenerate them. It's either a manual step or a nightly automated action.

What is your goal with this change?

PR #482, opened 2025-02-12
Allow DB override via hidden parameter

mkysel:
This is required to upgrade testing to create a separate DB that is only being used by the test.

Will get backported into a 0.1.3 stream.

## Summary by CodeRabbit

- **New Features**
	- Added a configuration option that allows you to override the default naming for your database setup, offering enhanced flexibility and customization of your database namespace.


fbac (approved):
LGTM. This can also be useful for reorg tests.

PR #525, opened 2025-02-20
Some good logging

mkysel:
Needed these to determine what is going on in the network.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced logging for improved contextual error reporting during node operations and stream handling.
	- Updated processing routines to provide more detailed diagnostic information, aiding in troubleshooting.


fbac (approved):
Good logging is immediately approved always.

PR #544, opened 2025-02-24
Verifiable payer

### LINKED ISSUE #474: Trust Violation - A malicious node can "steal" payloads
A node N can duplicate a payload from a different node N' proposed by payer P.

It can do so by taking the entire P payload and assigning it a new local sequenceID.

This would mean that the payer gets charged twice and a node can get paid for messages it was not meant to originate.

This needs to be fixed by allowing the payer to sign the `target_node_id`

mkysel:
Uses https://github.com/xmtp/proto/pull/247

This adds a new field to the payer envelope that prevents a node from stealing payloads https://github.com/xmtp/xmtpd/issues/474

Additionally it gets rid of old AAD validation #463

The bulk of the changeset allows the payer to pick a different node if the connection to the ideal node has failed. This is obviously not ideal, because it will keep retrying even if the node is down long term, but in the short/medium term this is sufficient.

Fixes #474 #463

## Summary by CodeRabbit

- **New Features**
  - Introduced a retry mechanism for publishing messages that dynamically selects alternative nodes upon failure.
  - Enhanced node selection to automatically exclude unavailable nodes, improving message routing reliability.
  - Added a new field for `TargetOriginator` in payer envelopes to improve message association.

- **Improvements**
  - Strengthened message validation and signature verification for better envelope integrity.
  - Streamlined error handling, resulting in clearer feedback in messaging workflows.
  - Enhanced test coverage for scenarios involving banned nodes.

- **Refactoring**
  - Unified parameter usage in envelope creation to ensure consistency across message processing.


pkg/api/payer/nodeSelector.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 type NodeSelectorAlgorithm interface {
-	GetNode(topic topic.Topic) (uint32, error)
+	GetNode(topic topic.Topic, banlist ...[]uint32) (uint32, error)
```
neekolas, line 13: Makes sense

neekolas (approved):

PR #557, opened 2025-02-26
Add initial Rates contract

neekolas:


neekolas:
* **#557** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/557?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/557?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

contracts/src/RatesManager.sol
```diff
@@ -0,0 +1,93 @@
...
+    // All Rates appended here
+    Rates[] private allRates;
+
+    // Event emitted when new Rates are added
+    event RatesAdded(uint64 messageFee, uint64 storageFee, uint64 congestionFee, uint64 startTime);
+
+    constructor()
+        AccessControlDefaultAdminRules(
+            1 days, // adminTransferDelay
+            msg.sender // initialAdmin
```
fbac, line 27: nit: In some scenarios the initialAdmin is different than the deployer. We could save the gas of transferring the role by having an `_initialAdmin` passed.

contracts/src/RatesManager.sol
```diff
@@ -0,0 +1,93 @@
...
+
+        // Grant the deployer the RATES_ADMIN_ROLE
+        _grantRole(RATES_ADMIN_ROLE, msg.sender);
+    }
+
+    /**
+     * @dev Add new Rates. Can only be called by addresses with RATES_ADMIN_ROLE.
+     *      The array only grows; we do not allow removal or updating.
+     */
+    function addRates(uint64 messageFee, uint64 storageFee, uint64 congestionFee, uint64 startTime)
```
fbac, line 41: Is there any legal reason/concern regarding having lower/upper bounds for fees?

contracts/src/RatesManager.sol
```diff
@@ -0,0 +1,93 @@
...
+     * @return hasMore - True if there are more items beyond this slice.
+     */
+    function getRates(uint256 fromIndex) external view returns (Rates[] memory rates, bool hasMore) {
+        if (allRates.length == 0 && fromIndex == 0) {
+            return (new Rates[](0), false);
+        }
+
+        require(fromIndex < allRates.length, "fromIndex out of range");
+
+        uint256 pageSize = 50; // Fixed page size
```
fbac, line 71: We could pass the offset as part of the call for flexibility.

If we still want to have a fixed page size, making this a constant would make the function cheaper; constants are calculated at compile time and we wouldn't have to store anything in memory.

contracts/src/RatesManager.sol
```diff
@@ -0,0 +1,93 @@
...
+    // Rates struct holds the fees and the start time of the rates
+    struct Rates {
+        uint64 messageFee;
+        uint64 storageFee;
+        uint64 congestionFee;
+        uint64 startTime;
+    }
+
+    // All Rates appended here
+    Rates[] private allRates;
```
fbac, line 19: This gives us ordering, but on the other hand is pretty expensive when we're looking for an specific timestamp in the past.

What about a `mapping(uint64 => Rates) private ratesByTimestamp;` and controlling wether to add the next timestamp with a `_lastTimestampAdded` or a cronological array `uint64[] private timestamps;`?
That would enable us to have a `getRateAt(timestamp)`

Additionally, the array is unbounded, we could consider pruning older timestamps when adding new ones, so the array doesn't grow beyond a certain point.

contracts/src/RatesManager.sol
```diff
@@ -0,0 +1,93 @@
...
+     */
+    function addRates(uint64 messageFee, uint64 storageFee, uint64 congestionFee, uint64 startTime)
+        external
+        onlyRole(RATES_ADMIN_ROLE)
+    {
+        // Enforce chronological order
+        require(
+            allRates.length == 0 || startTime > allRates[allRates.length - 1].startTime,
+            "startTime must be greater than the last startTime"
+        );
```
fbac, line 49: nit: I always default to revert with pre-defined errors, which can be easily parsed using ABIs, like InvalidStartTime() on this case.

contracts/src/RatesManager.sol
neekolas: My plan is to ingest all of them into the node and put them in a nice in-memory data structure there.

We do need all historical rates so that new nodes can get the correct Rates when processing very old messages.

contracts/src/RatesManager.sol
neekolas: Maybe. I'm waiting for more guidance on that, and won't put any bounds in until I'm sure it's necessary.

contracts/src/RatesManager.sol
neekolas: Good call. Will fix in the upstack PR

contracts/src/RatesManager.sol
neekolas: Good call

neekolas: Fixed in the upstack PR

contracts/src/RatesManager.sol
neekolas: This got fixed in the upstack PR as part of the move to upgradeable contracts

fbac (approved):

neekolas:
### Merge activity

* **Mar 4, 11:50 AM PST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/xmtp/xmtpd/557).

PR #558, opened 2025-02-26
feat: Payer interface

fbac:
**What's in this interface**
Core funcionality:
- Payer's account lifecycle.
- Deposits management, 2-steps withdraw, donations.
- Settlements.

Security:
- Pausability.
- Role based access.

Observability functions.

## Summary by CodeRabbit

- **New Features**
  - Introduced a comprehensive system for managing fund deposits, withdrawals, and usage tracking, enhancing secure financial operations.
  - Enhanced transparency with detailed event logging for deposits, withdrawals, and usage settlements.
  - Added administrative controls to manage thresholds and operational settings, ensuring smooth and secure interactions.
  - New structures and functions for managing payer accounts, including registration, balance management, and withdrawal processes.
  - Implemented error handling for various operational scenarios to improve user experience.

### DIFF
--- a/contracts/src/interfaces/IPayer.sol
+++ b/contracts/src/interfaces/IPayer.sol
@@ -1,0 +1,601 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/**
+ * @title IPayer
+ * @notice Interface for managing payer USDC deposits, usage settlements,
+ *         and a secure withdrawal process with optimized storage using
+ *         batch hash commitments and EIP-1283 gas optimizations.
+ */
+interface IPayer {
+    //==============================================================
+    //                             STRUCTS
+    //==============================================================
+
+    /**
+     * @dev Struct to store payer information.
+     * @param balance The current USDC balance of the payer.
+     * @param isActive Indicates whether the payer is active.
+     * @param creationTimestamp The timestamp when the payer was first registered.
+     * @param latestDepositTimestamp The timestamp of the most recent deposit.
+     * @param debtAmount The amount of fees owed but not yet settled.
+     */
+    struct Payer {
+        uint256 balance;
+        bool isActive;
+        uint256 creationTimestamp;
+        uint256 latestDepositTimestamp;
+        uint256 debtAmount;
+    }
+
+    /**
+     * @dev Struct to store withdrawal request information.
+     * @param requestTimestamp The timestamp when the withdrawal was requested.
+     * @param withdrawableTimestamp The timestamp when the withdrawal can be finalized.
+     * @param amount The amount requested for withdrawal.
+     */
+    struct WithdrawalRequest {
+        uint256 requestTimestamp;
+        uint256 withdrawableTimestamp;
+        uint256 amount;
+    }
+
+    /**
+     * @dev Struct to store usage reporting data.
+     * @notice This struct is used primarily for event emission and off-chain tracking.
+     * @dev For storage efficiency, complete reports are NOT stored on-chain; only their
+     *      cryptographic hash commitments are stored.
+     * @param payer The address of the payer being charged.
+     * @param nodeId The ID of the node that provided service.
+     * @param fees The amount charged for this usage.
+     * @param timestamp When the usage occurred.
+     */
+    struct UsageReport {
+        address payer;
+        uint256 nodeId;
+        uint256 fees;
+        uint256 timestamp;
+    }
+
+    /**
+     * @dev Struct to store batch commitment information.
+     * @param batchHash Hash of the complete batch data (payers, fees, timestamp, nodeId).
+     * @param totalFees The total fees collected in this batch.
+     * @param nodeId The node operator who submitted this batch.
+     * @param timestamp When this batch was processed.
+     * @param blockNumber The block in which this commitment was stored.
+     */
+    struct BatchCommitment {
+        bytes32 batchHash;
+        uint256 totalFees;
+        uint256 nodeId;
+        uint256 timestamp;
+        uint256 blockNumber;
+    }
+
+    //==============================================================
+    //                             EVENTS
+    //==============================================================
+
+    /// @dev Emitted when a new payer is registered.
+    event PayerRegistered(address indexed payer, uint256 amount);
+
+    /// @dev Emitted when a payer is deactivated by an owner.
+    event PayerDeactivated(address indexed payer);
+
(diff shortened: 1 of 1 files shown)


contracts/src/interfaces/IPayer.sol
```diff
@@ -0,0 +1,598 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
```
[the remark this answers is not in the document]
neekolas: All our other ones lock to 0.8.28

contracts/src/interfaces/IPayer.sol
```diff
@@ -0,0 +1,598 @@
...
+        address payer;
+        uint256 nodeId;
+        uint256 messageCount;
+        uint256 dataSize;
+        uint256 fees;
+        uint256 timestamp;
+    }
+
+    /**
+     * @dev Struct to store batch commitment information.
```
neekolas, line 65: Can you add a little more context around how batch commitments work

contracts/src/interfaces/IPayer.sol
```diff
@@ -0,0 +1,598 @@
...
+        uint256 lastSettlementTime
+    );
+
+    /**
+     * @notice Returns a list of payers with outstanding debt.
+     * @param limit Maximum number of payers to return.
+     * @return debtors Array of payer addresses with debt.
+     * @return debtAmounts Corresponding debt amounts for each payer.
+     */
+    function getPayersInDebt(uint256 limit) external view returns (
```
neekolas, line 516: Do we also want an offset to go with the limit?

contracts/src/interfaces/IPayer.sol
```diff
@@ -0,0 +1,598 @@
...
+    /**
+     * @notice Called by node operators to settle usage and calculate fees owed.
+     * @dev This function is EIP-1283 optimized by using accumulators for multiple state updates.
+     * @param fees The total USDC fees computed from this usage period.
+     * @param payer The address of the payer being charged.
+     * @param nodeId The ID of the node operator submitting the usage.
+     * @param messageCount Number of messages processed during this period.
+     * @param dataSize Total size of data processed in bytes.
+     * @param timestamp The timestamp when the usage occurred (can be backdated).
+     * @param commitmentHash Optional hash for additional off-chain data verification.
```
neekolas, line 336: What would the `commitmentHash` typically be set to?

contracts/src/interfaces/IPayer.sol
```diff
@@ -0,0 +1,598 @@
...
+
+    /**
+     * @notice Allows anyone to donate USDC to an existing payer's account.
+     *         The sender must approve this contract to spend USDC beforehand.
+     * @param payer The address of the payer receiving the donation.
+     * @param amount The amount of USDC to donate.
+     *
+     * Emits `DonationMade`.
+     */
+    function donate(address payer, uint256 amount) external;
```
neekolas, line 214: I like that idea

contracts/src/interfaces/IPayer.sol
fbac: So this is an idea aiming to make the protocol auditable from end to end. Completely optional, as we rely on trusted NOPs and the PayerReport attestation process.

We'd submit a PayerReport to the Treasury contract. After being accepted, Treasury calls payer.settleUsageBatch([]payers, []fees) which in turn creates a BatchCommitment storing hash(payers, fees, timestamp), subtract all the funds, etc.
The hash could be used to verify the settlement was processed correctly and noone tampered with the amounts/payers.

contracts/src/interfaces/IPayer.sol
fbac: This is a typo, thanks for catching it. The commitmentHash should be calculated by the settleUsage itself and not passed as param (security risk).

As mentioned before, the hash can be hash(payer, fee, nodeId, timestamp/block), any combination that gives us unique hashes and be used for verification. It's optional given our current design and we can't live without it.

neekolas: It does seem like we want to somehow guarantee uniqueness of the report, so I'm in favour. But yes, let's have the smart contract handle the hashing.

contracts/src/interfaces/IPayer.sol
```diff
@@ -0,0 +1,600 @@
...
+     * @notice Called by node operators to settle usage and calculate fees owed.
+     * @dev This function is EIP-1283 optimized by using accumulators for multiple state updates.
+     * @param fees The total USDC fees computed from this usage period.
+     * @param payer The address of the payer being charged.
+     * @param nodeId The ID of the node operator submitting the usage.
+     * @param timestamp The timestamp when the usage occurred (can be backdated).
+     *
+     * Emits `UsageSettled`.
+     */
+    function settleUsage(
```
neekolas, line 333: How do we make sure that `settleUsage` is not called more than once for the same payer report? Or is that the responsibility of another contract?

contracts/src/interfaces/IPayer.sol
```diff
@@ -0,0 +1,600 @@
...
+        uint256 fees,
+        address payer,
+        uint256 nodeId,
+        uint256 timestamp
+    ) external;
+
+    /**
+     * @notice Called by node operators to settle usage for multiple payers in a batch.
+     * @dev Uses EIP-1283 optimizations for storage efficiency and a simple hash commitment
```
neekolas, line 342: Is that the right EIP?

contracts/src/interfaces/IPayer.sol
neekolas: I see. So it's basically a "receipt" that the settlement was done.

contracts/src/interfaces/IPayer.sol
fbac: That would be responsibility of the contract handling Payer Reports, and settleUsage or settleUsageBatch can only be called from that contract.

contracts/src/interfaces/IPayer.sol
```diff
@@ -0,0 +1,600 @@
...
+     */
+    function deletePayer(address payer) external;
+
+    /**
+     * @notice Allows a payer to delete their own account from the system.
+     * @dev Can only be called if the payer has zero balance and zero debt and is not in withdrawal.
+     *
+     * Emits `PayerDeleted`.
+     */
+    function deleteMyAccount() external;
```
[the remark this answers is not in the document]
fbac: Deleted in the last iteration.

neekolas (approved):

Issue #470, opened 2025-02-10
Contracts: Rewards contract


fbac:
Rewards contract is responsible of distributing rewards across all the node operators and stakers of the protocol.

Issue #563, opened 2025-02-27
Merkleize Payer Reports?


neekolas:
One nagging concern with Payer Reports is their potentially unbounded size. If we want a report that covers 1,000,000 messages there can theoretically be 1,000,000 payers each with a minuscule balance change. 

Large reports will hit transaction size limits when submitted, and gas limits when we attempt to settle them all at once. 

Merkle Trees offer the prospect of us having Payer Reports of unlimited size without hitting gas or transaction limits.

## How it would work

1. When Nodes create a Payer Report they would create a Merkle Tree where each leaf represents the hash of a blob containing the payer address, their total spend in the report period, and any other metadata we want to store (maybe we want the total number of messages sent by the payer or other metadata to be used for rebate calculations)
2. The onchain payer report only contains the Merkle Root
3. The report gets enough attestations that it can be confirmed
4. Anyone can settle the usage of an individual payer (or batch of payers) by submitting the original blob and a Merkle Proof showing its inclusion in the tree. This allows us to break up a very large report into many transactions to settle if needed.

## Challenges

We need to make sure that an individual payer's component of the report cannot be settled more than once. That requires some extra bookkeeping.

Some ideas:
- We maintain a list of hashes that have already been settled onchain, and then eventually delete that list when the report is completely paid out. Easiest...but adds on-chain storage cost. Still effectively unlimited in size because it allows for settlement to be broken into multiple transactions.
- We use a [Sparse Merkle Tree](https://github.com/kevincharm/sparse-merkle-tree) instead of a regular Merkle Tree, and provide "update proofs" when settling individual payers. This allows the Merkle Root to change each time a set of payers is settled. Would need to figure out how the caller can know which nodes have been removed so that they can provide subsequent update proofs based on the current tree.
- Maybe there is some way to have a settlement remove an entire branch of the tree (some number of payer blobs) and replace the Merkle Root with a new hash representing the remaining branches. Would need to be provable that the new Root represents the remainder of the tree with nothing added or removed. Not sure if there is a version of Merkle Trees that supports this. Sparse Merkle Trees are close, but the updates have to be one node at a time as far as I can tell.
- Something else...?

neekolas:
OK. I think I have an idea that is workable on paper.

[...]

PR #581, opened 2025-03-03
IPayerReport interface

fbac:
Context around Merkle Trees in https://github.com/xmtp/xmtpd/issues/563
Closes https://github.com/xmtp/xmtpd/issues/470?issue=xmtp%7Cxmtpd%7C499

The following is the expected sequence of events:
- A node generates and submits a `PayerReport`.
- All the other nodes call `attestPayerReport` if they can confirm its validity.
- Nodes in the network listen for attestations and tracks them. Upon reaching majority, the `confirmPayerReport` can be called.
- When a `PayerReport` is confirmed, any node can submit batches to `settleUsageBatch`. The `PayerReport` contract is expected to track batches submissions to prevent duplicates.
- `settleUsageBatch` verifies the provided payers and amounts leafs are included in the PayerReport Merkle tree. If verified, the Payer.settleUsage is called with the list of payers and amounts to settle the usage. 

Nodes attest to PayerReports, and once majority has been reached `confirmPayerReport` can be called for settlement.

## Summary by CodeRabbit

- **New Features**
  - Enhanced the usage settlement process to handle batches more efficiently with improved verification.
  - Introduced a new reporting interface enabling users to submit, attest, and confirm usage data reports.
  - Added events for tracking report submissions, attestations, and confirmations.
  - Updated terminology for contracts related to distribution to better reflect current functionality.
- **Bug Fixes**
  - Clarified documentation for events to accurately represent their context and usage.

### DIFF
--- a/contracts/src/interfaces/IPayer.sol
+++ b/contracts/src/interfaces/IPayer.sol
@@ -70,9 +70,6 @@ interface IPayer {
     /// @dev Emitted when usage is settled and fees are calculated.
     event UsageSettled(uint256 fees, address indexed payer, uint256 indexed nodeId, uint256 timestamp);
 
-    /// @dev Emitted when batch usage is settled.
-    event BatchUsageSettled(uint256 totalFees, uint256 indexed nodeId, uint256 timestamp);
-
     /// @dev Emitted when fees are transferred to the rewards contract.
     event FeesTransferred(uint256 amount);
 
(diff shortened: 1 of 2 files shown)


contracts/src/interfaces/IPayer.sol
```diff
@@ -264,54 +261,37 @@ interface IPayer {
...
-     * @dev Uses EIP-2200 optimizations for storage efficiency.
-     * @param payers Array of payer addresses being charged.
-     * @param fees Array of USDC fees corresponding to each payer.
-     * @param timestamp When this batch of usage occurred (can be backdated).
-     * @param nodeId The ID of the node operator submitting the usage.
-     *
-     * Emits `BatchUsageSettled` and multiple `UsageSettled` events.
-     */
-    function settleUsageBatch(
+        address originatorNode,
```
neekolas, line 274: I believe this will also need to include a Batch Merkle Proof, won't it?

neekolas:
@fbac 
> Upon reaching majority, the confirmPayerReport can be called.

Could we do this automatically as the attestations come in. Otherwise I'm not sure who would be responsible for calling this

contracts/src/interfaces/IPayerReport.sol
```diff
@@ -0,0 +1,162 @@
...
+    //==============================================================
+
+    struct UsageReport {
+        address originatorNode;
+        uint256 startingSequenceID;
+        uint256 endingSequenceID;
+        uint256 lastMessageTimestamp;
+        uint256 reportTimestamp;
+        bytes32 reportMerkleRoot;
+        address[] payers;
```
neekolas, line 20: With the merkle root, would we actually need the `payers` and `amountsSpent` on this struct?

contracts/src/interfaces/IPayerReport.sol
```diff
@@ -0,0 +1,162 @@
...
+     */
+    event PayerReportSubmitted(
+        address indexed originatorNode,
+        uint256 indexed reportIndex,
+        bytes32 indexed reportMerkleRoot,
+        uint256 startingSequenceID,
+        uint256 endingSequenceID,
+        uint256 lastMessageTimestamp,
+        uint256 reportTimestamp,
+        address[] payers,
```
neekolas, line 42: Same question as above

contracts/src/interfaces/IPayerReport.sol
```diff
@@ -0,0 +1,162 @@
...
+        bytes32 indexed reportMerkleRoot
+    );
+
+    /**
+     * @dev Emitted when a batch of usage is settled.
+     */
+    event UsageSettled(
+        address indexed originatorNode,
+        uint256 indexed reportIndex,
+        uint256 offset,
```
neekolas, line 70: For this event to be indexable by the nodes I believe we would need to have the payers and the amounts

neekolas: Otherwise the node would have to do a bunch of work to reconstruct the tree and figure out what was being settled.

fbac:
> @fbac
> 
> > Upon reaching majority, the confirmPayerReport can be called.
> 
> Could we do this automatically as the attestations come in. Otherwise I'm not sure who would be responsible for calling this

@nick I was concerned about having a clear separation of concerns in the contract ABI, so attestPayerReport doesn't return 2 types of events depending the amount of confirmations, and Nodes would attest to a PayerReport while counting offchain the confirmations, then any node could call confirmPayerReport.

But yes, it's definitely possible to combine the logic into `attestPayerReport`.

contracts/src/interfaces/IPayerReport.sol
fbac: That's correct, we don't. Each node can compute the lists off-chain.

contracts/src/interfaces/IPayer.sol
fbac: Actually I think offset is not needed in Payer.settleUsage function.
The way I see it is the PayerReport.settleUsageBatch accepts the proofs and verifies them. Only when the []payers and []amounts are verified as part of the merkle tree we call Payer.settleUsage([]payer, []amount), which is a dumb function just to settle what we agreed on.

contracts/src/interfaces/IPayerReport.sol
fbac: Agree, modifying this part too.

Also, with the latest commit I added a `uint16 batches` to the PayerReport struct, which should be calculated off-chain based on `getMaxBatchSize()`. Other nodes can reconstruct that data and agree on it.

I think this is interesting because we don't submit the amount of payers and amounts anymore, so I don't see a reliable way of calculating the amount of batches expected for a single report, tracking what was already settled, etc.

neekolas: This makes sense to me. Do we want it to be the number of batches, or just the number of leaves in the tree? I'm fine with either, but the number of leaves lets the caller adjust the batch size as needed.

neekolas (approved):

contracts/src/interfaces/IPayerReport.sol
fbac: I like the number to be leaves in the tree. It reduces ambiguity and hidden knowledge (such as batches has to be N payers max)

PR #589, opened 2025-03-05
Breaking: Integrate NodesV2

fbac:
Summary
--
- xmtpd now uses `NodesV2`. **This is a breaking change!**.
- Payer node selection skips nodes disabled (enforced by admins) or with not enabled api (enforced by NOPs)
- Sync stops subscribing to disabled nodes (enforced by admins) or with replication disabled (enforced by NOPs)
- Minor changes: fixed some error strings (linter complains about them starting with uppercase) and removed an unused interface.


pkg/api/payer/nodeSelector.go
```diff
@@ -53,9 +54,19 @@ func (s *StableHashingNodeSelectorAlgorithm) GetNode(
...
+		if node.IsDisabled || !node.IsApiEnabled {
+			continue
+		}
 		nodeLocations[i] = uint32(i) * spacing
+		gotNodes = true
+	}
+
+	if !gotNodes {
+		return 0, errors.New("no available active nodes")
 	}
```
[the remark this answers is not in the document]
fbac: With the current approach using gotNodes we don't need a new for loop, saving cpu cycles.

pkg/mocks/registry/mock_NodesContract.go
```diff
@@ -22,24 +22,24 @@ func (_m *MockNodesContract) EXPECT() *MockNodesContract_Expecter {
 	return &MockNodesContract_Expecter{mock: &_m.Mock}
 }
 
-// AllNodes provides a mock function with given fields: opts
-func (_m *MockNodesContract) AllNodes(opts *bind.CallOpts) ([]nodes.NodesNodeWithId, error) {
+// GetAllNodes provides a mock function with given fields: opts
+func (_m *MockNodesContract) GetAllNodes(opts *bind.CallOpts) ([]nodesv2.INodesNodeWithId, error) {
```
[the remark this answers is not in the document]
fbac: This will be updated in following PRs.

pkg/registry/node.go
```diff
@@ -13,11 +14,14 @@
...
-	IsValidConfig bool
+	NodeID               uint32
+	SigningKey           *ecdsa.PublicKey
+	HttpAddress          string
+	IsReplicationEnabled bool
+	IsApiEnabled         bool
+	IsDisabled           bool
+	MinMonthlyFee        *big.Int
+	IsValidConfig        bool
 }
```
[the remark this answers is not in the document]
fbac: This will be updated in following PRs.

pkg/registry/interface.go
```diff
@@ -2,15 +2,15 @@ package registry
...
+	"github.com/xmtp/xmtpd/contracts/pkg/nodesv2"
 )
 
 /*
 *
 A dumbed down interface of abis.NodesCaller for generating mocks
 */
 type NodesContract interface {
-	AllNodes(opts *bind.CallOpts) ([]nodes.NodesNodeWithId, error)
+	GetAllNodes(opts *bind.CallOpts) ([]nodesv2.INodesNodeWithId, error)
```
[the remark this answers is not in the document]
fbac: This will be updated in following PRs.

pkg/api/payer/nodeSelector.go
```diff
@@ -53,9 +54,19 @@ func (s *StableHashingNodeSelectorAlgorithm) GetNode(
 	spacing := maxHashSpace / numNodes
 
 	// Compute virtual positions for each node
+	// Skip nodes that are disabled or do not have an API enabled
 	nodeLocations := make([]uint32, numNodes)
```
neekolas, line 58: If this is the length of `numNodes` but some are being skipped in the loop below the array is going to have empty values.

Better to use `append` and calculate `numNodes` based on the filtered list.

pkg/api/payer/nodeSelector.go
```diff
@@ -53,9 +54,19 @@ func (s *StableHashingNodeSelectorAlgorithm) GetNode(
...
-	for i := range nodes {
+	gotNodes := false
+	for i, node := range nodes {
+		if node.IsDisabled || !node.IsApiEnabled {
+			continue
+		}
 		nodeLocations[i] = uint32(i) * spacing
```
neekolas, line 64: I see your conversation with the AI below about the extra for loop, but I think having empty indexes in this array is going to be a footgun that will make the code harder to reason about. 

Maybe there is a third solution where we can use a single loop without having empty indexes in the `nodeLocations`?

neekolas (approved):

pkg/api/payer/nodeSelector.go
fbac: It shouldn't apply here in the payer. The contract registry gets rid of those empty nodes during `refreshData()`, right?
```
	newNodes := []Node{}
	for _, node := range fromContract {
		// nodes realistically start at 100, but the contract fills the array with empty nodes
		if !node.IsValidConfig {
			continue
		}
```

neekolas: Maybe we're taking about different things. We have a `nodeLocations` array of length `numNodes`, but only the indexes with `IsDisabled = false && IsApiEnabled = true` have data in them. The rest of the indexes are set to 0 because of the `continue` on line 62.

fbac: Oh, I see. Good catch! I just modified a bit the test `TestGetNode_NoAvailableActiveNodes` and triggered an error because of this. Modifying this.

neekolas (approved):

pkg/api/payer/nodeSelector.go
fbac: Modified. I moved the banlist filter up as well, now we filter earlier, sort, and return the candidate.

pkg/api/payer/nodeSelector.go
```diff
@@ -39,22 +40,46 @@ func (s *StableHashingNodeSelectorAlgorithm) GetNode(
...
 	}
 
-	// Sort nodes to ensure stability
-	sort.Slice(nodes, func(i, j int) bool { return nodes[i].NodeID < nodes[j].NodeID })
+	// Filter out banned, disabled, or non-API nodes
+	var availableNodes []registry.Node
+	for _, node := range nodes {
+		if node.IsDisabled || !node.IsApiEnabled {
+			continue
+		}
```
mkysel, line 56: I don't think that this is the right place to do this. reg.GetNodes() already does a bunch of pre-filtering. Why dont we do it there so its all in one place? Look at `func (s *SmartContractRegistry) refreshData() error`

pkg/api/payer/nodeSelector.go
```diff
@@ -39,22 +40,46 @@ func (s *StableHashingNodeSelectorAlgorithm) GetNode(
...
+		if _, exists := banned[node.NodeID]; !exists {
```
neekolas, line 57: If the number of available nodes changes based on the ban list, then the partitions will change every time you add a member to the banlist and retry.

I don't think that works. The consistent hashing ring should be consistent between retries.

pkg/registry/contractRegistry.go
```diff
@@ -47,14 +47,17 @@ type SmartContractRegistry struct {
 	cancel                    context.CancelFunc
 }
 
+// Interface implementation guard.
+var _ NodeRegistry = &SmartContractRegistry{}
```
mkysel, line 51: ooh. I didn't know about this. Thanks for teaching me!

mkysel (approved):

pkg/api/payer/nodeSelector.go
neekolas: What if we did something like this, so that the `numNodes` and spacing doesn't change unless the result set from the registry changes.

```go
// GetNode selects a node for a given topic using stable hashing
func (s *StableHashingNodeSelectorAlgorithm) GetNode(
	topic topic.Topic,
	banlist ...[]uint32,
) (uint32, error) {
	nodes, err := s.reg.GetNodes()
	if err != nil {
		return 0, err
	}

	// Flatten banlist
	banned := make(map[uint32]struct{})
	for _, list := range banlist {
		for _, id := range list {
			banned[id] = struct{}{}
		}
	}

	// Sort availableNodes to ensure stability
	sort.Slice(nodes, func(i, j int) bool {
		return nodes[i].NodeID < nodes[j].NodeID
	})

	topicHash := HashKey(topic)

	numNodes := uint32(len(nodes))
	maxHashSpace := ^uint32(0)
	spacing := maxHashSpace / numNodes

	// Compute virtual positions for each available node
	nodeLocations := make([]uint32, numNodes)
	for i := range availableNodes {
		nodeLocations[i] = uint32(i) * spacing
	}

	// Binary search to find the first node with a virtual position >= topicHash
	idx := sort.Search(len(nodeLocations), func(i int) bool {
		node := nodes[i]
		if node.IsDisabled || !node.IsApiEnabled {
			return false
		}
		if _, banned := banned[node.NodeID]; banned {
			return false
		}
		return topicHash < nodeLocations[i]
	})

	// Select the appropriate node from availableNodes
	candidateIdx := idx % len(nodeLocations)
	return availableNodes[candidateIdx].NodeID, nil
}
```

Issue #486, opened 2025-02-14
Improve upgrade tests by using testcontainers-go


mkysel:
The poor mans solution just execs on the host. Use a proper go library instead.

Suggested by @fbac

Issue #560, opened 2025-02-27
Improve upgrade tests by managing contracts state


fbac:
Problem statement
--
Currently the upgrade test don't deploy or upgrade contracts, instead they rely on the blockchain status blindly. 
If an upgrade requires new contracts to be deployed, or an old ABI has changed, the upgrade test fails.

Expected behavior
--
Upgrade test is able to deploy, upgrade or re-deploy the needed contracts introduced on each version.
A naive flow would be:
```
old version test start
- deploy needed contracts.
- load envVars.
- test

new version test start
- deploy new contracts if any.
- upgrade existing contracts if there's an upgrade script available.
- load envVars again, as an old envVar could be overridden by a new contract address.
- test
```

PR #603, opened 2025-03-10
Use testcontainers in upgrade tests

fbac:
Closes https://github.com/xmtp/xmtpd/issues/486
- Simplify upgrade test logic by using testcontainers.
- Continue on error in upgrade tests, instead make a comment with test results.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new GitHub Actions workflow for automated upgrade testing.
- **Chores**
  - Updated multiple dependencies to boost stability, performance, and security.
- **Tests**
  - Enhanced functionality to pull Docker images directly in tests.
  - Streamlined the upgrade test setup, including a dedicated test validating the latest development version.
  - Improved error handling and context management in test functions.
  - Refined container-based tests for improved reliability and clarity.
  - Improved version management in upgrade tests for better structure and clarity.
  - Reordered import statements in test files for better organization.
  - Removed the "Run Upgrade Tests" step from the GitHub Actions workflow.


.github/workflows/test.yml
```diff
@@ -46,11 +46,28 @@ jobs:
           export PATH="${PATH}:${GOPATH}/bin"
           go test -v ./... -race
       - name: Run Upgrade Tests
+        id: upgrade_tests
+        continue-on-error: true
```
mkysel, line 50: why do we want to mark failing tests as passed?

pkg/upgrade/main_test.go
```diff
@@ -22,18 +22,6 @@ func setup() {
...
-	pullStart := time.Now()
-	for _, image := range upgradeToLatest {
-		err := dockerPull(image)
-		if err != nil {
-			fmt.Printf("    ❌ Error pulling image %s: %v\n", image, err)
-			os.Exit(1)
-		}
-	}
-	fmt.Printf("    ✔ All images pulled in %v\n", time.Since(pullStart))
-
```
mkysel, line 36: this will cause flakiness. You really don't want to pull the images on demand in a test. Random tests might time out

.github/workflows/test.yml
mkysel: comments are extremely easy to miss. Failed tests less so

pkg/upgrade/docker_utils_test.go
```diff
@@ -96,91 +89,6 @@ func constructVariables(t *testing.T) map[string]string {
...
-	var dockerEnvArgs []string
-	for key, value := range envVars {
-		dockerEnvArgs = append(dockerEnvArgs, "-e", fmt.Sprintf("%s=%s", key, value))
-	}
-
-	_ = dockerRmc(containerName)
-
-	dockerCmd := []string{"run", "-d"}
-	if runtime.GOOS == "linux" {
-		dockerCmd = append(dockerCmd, "--add-host=host.docker.internal:host-gateway")
```
mkysel, line 142: you need this line

.github/workflows/test.yml
fbac: You're right, looking at it I don't like this approach either.
The original idea was to separate go unit testing from upgrade testing, as they serve different purposes, so we can know what exactly is failing without having to open the gh action. Instead of the current workaround I'll be splitting it into a different workflow, so we don't see continuous "go test failing" when it's an upgrade test.

pkg/upgrade/docker_utils_test.go
fbac: Yeah, realized that when it ran in an ubuntu container. Added into as part of the ContainerRequest.

mkysel:
I suspect that it does fix #486, but we still need to fix #560 before the tests turn green again

fbac:
> I suspect that it does fix #486, but we still need to fix #560 before the tests turn green again

That's correct, this PR won't fix the contracts divergence. On the other hand, a contract ABI change as big as this one (Nodes V1 to V2) is 100% an upgrade breaking, so it's expected.

On the other hand, this could be a point in favor of splitting the contracts and xmtpd repositories, and have different versioning on both. From xmtpd perspective we only care of the contracts ABI at a specific version, from the contract perspective the development should continue without breaking things in xmtpd.
After that a compatibility matrix could be calculated. And in breaking upgrade events handle it with an specific procedure.

mkysel (approved):

PR #612, opened 2025-03-11
Enhance node filtering

fbac:
## Summary by CodeRabbit

- **Refactor**
	- Enhanced node selection to choose only fully enabled service nodes, ensuring more reliable operation and clearer error messaging.
	- Streamlined handling of node subscriptions by utilizing the full Node object for improved clarity and efficiency.

- **Tests**
	- Updated test procedures with standardized node simulations and added new scenarios verifying behavior when no eligible nodes are available.

### DIFF
--- a/pkg/api/payer/nodeSelector.go
+++ b/pkg/api/payer/nodeSelector.go
@@ -44,8 +44,10 @@ func (s *StableHashingNodeSelectorAlgorithm) GetNode(
 		return 0, errors.New("no available nodes")
 	}
 
-	// Sort nodes to ensure stability
-	sort.Slice(nodes, func(i, j int) bool { return nodes[i].NodeID < nodes[j].NodeID })
+	// Sort availableNodes to ensure stability
+	sort.Slice(nodes, func(i, j int) bool {
+		return nodes[i].NodeID < nodes[j].NodeID
+	})
 
 	topicHash := HashKey(topic)
 
@@ -75,12 +77,18 @@ func (s *StableHashingNodeSelectorAlgorithm) GetNode(
 	// Find the next available node
 	for i := 0; i < len(nodes); i++ {
 		candidateIdx := (idx + i) % len(nodeLocations)
-		candidateNodeID := nodes[candidateIdx].NodeID
+		node := nodes[candidateIdx]
 
-		if _, exists := banned[candidateNodeID]; !exists {
-			return candidateNodeID, nil
+		// Skip api disabled nodes. Admin disabled nodes are not returned by the registry.
+		// This does not affect the stability of the algorithm, is preserves the node order.
+		if !node.IsApiEnabled {
+			continue
+		}
+
+		if _, exists := banned[node.NodeID]; !exists {
+			return node.NodeID, nil
 		}
 	}
 
-	return 0, errors.New("no available nodes after considering banlist")
+	return 0, errors.New("no available nodes after filtering")
 }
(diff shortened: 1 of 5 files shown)


pkg/api/payer/nodeSelector.go
```diff
@@ -75,12 +77,18 @@ func (s *StableHashingNodeSelectorAlgorithm) GetNode(
 	// Find the next available node
 	for i := 0; i < len(nodes); i++ {
 		candidateIdx := (idx + i) % len(nodeLocations)
-		candidateNodeID := nodes[candidateIdx].NodeID
+		node := nodes[candidateIdx]
+
+		// Skip api disabled nodes. Admin disabled nodes are not returned by the registry.
+		// This does not affect the stability of the algorithm, is preserves the node order.
+		if !node.IsApiEnabled {
```
mkysel, line 84:
```{
  "level": "INFO",
  "time": "2025-03-11T16:01:28.456-0400",
  "message": "got nodes",
  "size": 2,
  "nodes": [
    {
      "node_id": 100,
      "owner_address": "0x8ECD607fB9f4e90e20a776D6F6EEfeCB87Cfda6a",
      "signing_key_pub": "0x028b98270cea6df37caab7126be932f52aac0fb56b4fafe9a91184c1b04b187084",
      "http_address": "https://grpc.testnet-playground.xmtp.network",
      "min_monthly_fee_micro_dollars": 0,
      "is_replication_enabled": false,
      "is_api_enabled": false
    },
    {
      "node_id": 200,
      "owner_address": "0x3e85013ED6f7eeC951E68dDF0C1f3f6dCCfAEAD3",
      "signing_key_pub": "0x03e5442c5d1fe2f02b6b9a1a386383a7766860b40a6079a0223994ffa2ce10512c",
      "http_address": "https://grpc2.testnet-playground.xmtp.network",
      "min_monthly_fee_micro_dollars": 0,
      "is_replication_enabled": false,
      "is_api_enabled": false
    }
  ]
}
```

all playground nodes have API enabled. Once this PR lands, all nodes will become unreachable. How does one turn them on?

pkg/sync/syncWorker.go
```diff
@@ -123,40 +123,45 @@ func (s *syncWorker) subscribeToRegistry() {
...
-	if nodeid == s.registrant.NodeID() {
+func (s *syncWorker) subscribeToNode(node registry.Node) {
+	if node.NodeID == s.registrant.NodeID() {
+		return
+	}
+
+	if !node.IsReplicationEnabled {
+		s.log.Debug("Skipping node not enabled for replication", zap.Uint32("nodeID", node.NodeID))
 		return
 	}
```
mkysel, line 142: I am still lost in the distinction between replicationEnabled and APIEnabled. We use these terms interchangeably. API refers to the `GRPC` API and replication refers to the V4 `message_api`, which happens to be exactly the same thing. For example `XMTPD_REPLICATION_ENABLE` enables the `message_api` GRPC endpoint.

Implementation wise a node does not distinguish between incoming node sync requests and incoming client requests. Eventually nodes can provide additional auth to be exempt from the rate limiter.

It seems that `IsApiEnabled` means "do not forward payer write requests to me" and `IsReplicationEnabled` means "do not sync from me".

It would be helpful to have a very clear naming scheme here AND it needs to be clearly documented, because we will be explaining this to our testnet partners over and over again. I don't want to have to read code every time someone asks "Do I have to enable replication?"

mkysel: to futher add to the confusion, one would logically assume that `IsReplicationEnabled` is related to `XMTPD_REPLICATION_ENABLE`. It is, if you set ReplicationEnabled to true but XMTPD_REPLICATION_ENABLE to false, all sync requests will fail. But so will all client requests.

PR #621, opened 2025-03-12
Do not skip storing messages in indexer

fbac:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling during event processing to ensure continuity even if blockchain queries fail, preventing unnecessary interruptions in operations.

### DIFF
--- a/pkg/indexer/indexer.go
+++ b/pkg/indexer/indexer.go
@@ -286,50 +286,54 @@ func indexLogs(
 		// - There's a stored block
 		// - The event block number is greater than the stored block number
 		// - The check interval has passed
+		skipReorgHandling := false
 		if !reorgInProgress &&
 			storedBlockNumber > 0 &&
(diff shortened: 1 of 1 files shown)


mkysel (approved):

pkg/indexer/indexer.go
```diff
@@ -286,50 +286,54 @@ func indexLogs(
 		// - There's a stored block
 		// - The event block number is greater than the stored block number
 		// - The check interval has passed
+		skipReorgHandling := false
 		if !reorgInProgress &&
```
mkysel, line 290: I have never managed to truly convince myself that this `if` statement is correct. I just glaze over it and say "this is too complicated for me to dig into". Not the best approach, I know. Adding another layer of if statements makes this virtually impossible to reason about. I am hoping that someone other than Borja knows what this does :)

PR #622, opened 2025-03-12
Add payer metrics

mkysel:
Far from complete, but a step in the right direction.


pkg/metrics/payer.go
```diff
@@ -0,0 +1,71 @@
...
+
+func EmitNodePublishDuration(originatorId uint32, duration float64) {
+	nodePublishDuration.With(prometheus.Labels{"originator_id": strconv.Itoa(int(originatorId))}).
+		Observe(duration)
+}
+
+var cursorBlockTime = prometheus.NewHistogramVec(
+	prometheus.HistogramOpts{
+		Name: "xmtp_payer_cursor_block_time_seconds",
+		Help: "Block time to reach the desired cursor",
```
neekolas, line 24: I want to make sure I understand this metric. What do we mean by block time?

mkysel: I am terrible at naming things. It shows you how long the payer blocked until the desired node reached the cursor. aka these steps:
- connected to node
- node has indexed the chain
- node has processed the new message
- node has made the message available
- payer confirmed that the message is available

This is required for read-your-own-commits

fbac: xmtp_read_own_commit_in_time_seconds?

Keeping it simple.

mkysel: love it. Can you give me the `help text` as well :D?

neekolas (approved):

PR #658, opened 2025-03-21
Add loadtest script

neekolas:


neekolas:
* **#658** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/658?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/658?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pkg/blockchain/nonceManager.go
```diff
@@ -51,7 +52,7 @@ func NewSQLBackedNonceManager(db *sql.DB, logger *zap.Logger) *SQLBackedNonceMan
 	return &SQLBackedNonceManager{
 		db:      db,
 		logger:  logger.Named("SQLBackedNonceManager"),
-		limiter: NewOpenConnectionsLimiter(BestGuessConcurrency),
+		limiter: NewOpenConnectionsLimiter(MaxConcurrentRequests),
```
neekolas, line 55: @mkysel is this change a bad idea?

pkg/testutils/noncemanager/nonceManager.go
```diff
@@ -0,0 +1,90 @@
+package noncemanager
```
neekolas, line 1: @mkysel I moved this to a test util when I was thinking about using it for the loadtest. But I was getting a lot of errors which all went away when I switched to the database-backed nonce manager. Maybe a cause of some test flakes?

pkg/testutils/noncemanager/nonceManager.go
```diff
@@ -0,0 +1,90 @@
...
+		nonce = tm.nonce
+		tm.nonce++
+	}
+
+	tm.logger.Debug("Generated Nonce", zap.Int64("nonce", nonce))
+
+	return &blockchain.NonceContext{
+		Nonce: *new(big.Int).SetInt64(nonce),
+		Cancel: func() {
+			tm.mu.Lock()
```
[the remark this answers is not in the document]
neekolas: @mkysel hmm. Does the AI have a point here?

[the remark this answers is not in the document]
mkysel: possibly. The `abandoned` test code is untested

pkg/blockchain/nonceManager.go
mkysel: you will break the deployed environment. 2x64 is too much if you hammer in HA. But it is worth exploring. I have not done any tuning.

pkg/loadtest/loadtest.go
```diff
@@ -0,0 +1,140 @@
...
+	elapsed := time.Since(startTime)
+	actualRate := float64(sentCount) / elapsed.Seconds()
+
+	logger.Info("Load test completed",
+		zap.Int("messagesSent", sentCount),
+		zap.Duration("elapsed", elapsed),
+		zap.Float64("actualMessagesPerSecond", actualRate))
+}
+
+func randomBytes(n int) []byte {
```
mkysel, line 129: we already have this in testutils :) as well as `RandomGroupID`

pkg/blockchain/nonceManager.go
neekolas: OK. I'll revert it for now and play around a bit

pkg/loadtest/loadtest.go
```diff
@@ -0,0 +1,140 @@
...
+		go func(messageNumber int) {
+			defer wg.Done()
+			if _, err := publisher.PublishGroupMessage(ctx, groupID, randomBytes(options.MessageSizeBytes)); err != nil {
+				logger.Error(
+					"Failed to publish message",
+					zap.Error(err),
+					zap.Int("messageNumber", messageNumber+1),
+				)
+			}
+		}(i)
```
mkysel, line 107: can we return an error code all the way to the client? That way you can do something along the lines of `while (loadtest); do...`

mkysel:
do you already have results from localhost? I wonder how different anvil is from say staging

neekolas:
I've only tested against testnet staging.

I don't think it's hitting staging at its full capacity though. When I look in the block explorer I don't see fees increasing, which would be a clear indicator that the blocks are full.

pkg/loadtest/loadtest.go
neekolas: I don't want to give up on a single error. But I could turn them into an array and return the whole thing when the test is done.

mkysel: either that or just return "has failed yes/no". I agree that early exit is undesired.

PR #683, opened 2025-03-30
Remove unsupported formatters from golangci-lint config

Fallengirl:
Remove gosimple from the enable section, as it is no longer supported in newer versions of golangci-lint. The other linters can be left. Without these changes, golangci-lint fails with errors. Locally tested per the [official migration guide.](https://golangci-lint.run/product/migration-guide/)

## Summary by CodeRabbit

- **Chores**
  - Made adjustments to internal code quality configurations to streamline analysis tools.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -7,7 +7,6 @@ linters:
     - nakedret
     - nilerr
     - errcheck
-    - gosimple
     - govet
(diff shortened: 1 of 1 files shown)


mkysel:
we are still on `golangci-lint` v1. This would reduce our coverage.

PR #694, opened 2025-04-02
Sync worker reliability

neekolas:
## tl;dr

- Moves a lot of the business logic into the `originatorStream`
- Adds a write queue to separate validating messages and writing to the DB
- Better handles retryable vs non-retryable errors. Anything that happens in the write queue is always retryable, anything that happens when processing a message from the stream is not.
- Refactors some of our test helpers to make mocking the SyncWorker easier

## Summary by CodeRabbit

- **New Features**
  - Introduced a new package for mock generation, enhancing testing capabilities.
  - Added a new field for target rate per minute in the fee structure, improving fee calculation functionalities.
  - Created functions for node and mock registry creation, facilitating easier testing setups.
  - Added comprehensive unit tests for the sync worker to validate envelope processing and error handling.

- **Bug Fixes**
  - Enhanced processing reliability by implementing an exponential backoff retry mechanism for envelope handling, improving error management and system stability.

- **Development Environment**
  - Enabled language server support and specified formatting preferences for Go code in the development environment settings.


neekolas:
* **#694** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/694?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/694?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.vscode/settings.json
```diff
@@ -12,6 +12,10 @@
     "reservedWordCase": "upper",
     "typeCase": "upper"
   },
+  "go.useLanguageServer": true,
+  "gopls": {
+    "formatting.gofumpt": true
```
neekolas, line 17: To make the default formatter in vscode `gofumpt`

pkg/testutils/registry/registry.go
```diff
@@ -0,0 +1,46 @@
...
+	mockRegistry.On("OnNewNodes").
+		Return((<-chan []r.Node)(nodesChan), r.CancelSubscription(func() {}))
+
+	for _, node := range nodes {
+		nodeChan := make(chan r.Node)
+		mockRegistry.On("OnChangedNode", node.NodeID).
+			Return((<-chan r.Node)(nodeChan), r.CancelSubscription(func() {
+				close(nodeChan)
+			}))
+		mockRegistry.On("GetNode", node.NodeID).Return(&node, nil)
```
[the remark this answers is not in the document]
neekolas: This is out of date with newer versions of go

pkg/sync/syncWorker.go
```diff
@@ -430,10 +452,11 @@ func (s *syncWorker) validateAndInsertEnvelope(
 		}
 	}()
 
-	env, err := envUtils.NewOriginatorEnvelope(envProto)
+	var env *envUtils.OriginatorEnvelope
+	env, err = envUtils.NewOriginatorEnvelope(envProto)
 	if err != nil {
 		s.log.Error("Failed to unmarshal originator envelope", zap.Error(err))
-		return err
+		return backoff.Permanent(err)
```
mkysel, line 459: this will shut down the worker. Which, given how complex this nesting is, will retry again here ```tracing.GoPanicWrap(
		s.ctx,
		&s.wg,
		fmt.Sprintf("node-subscribe-%d", nodeid),
		func(ctx context.Context) {
			for {
				select {
				case <-ctx.Done():
					return
				default:
					config := s.setupNodeRegistration(ctx, nodeid)
					s.subscribeToNodeRegistration(config)
				}
			}
		})```

mkysel: what should we do when we receive something we cant process?

mkysel:
@neekolas can you give me a bit of context around this? What prompted this changeset. Did you see something broken? How does this fix the issue.

pkg/sync/syncWorker.go
```diff
@@ -364,6 +365,9 @@ func (s *syncWorker) listenToStream(
 	recvChan := make(chan *message_api.SubscribeEnvelopesResponse)
 	errChan := make(chan error)
 
+	expBackoff := backoff.NewExponentialBackOff()
```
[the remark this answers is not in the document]
neekolas: I agree the nesting isn't ideal, but that backoff is serving a different purpose. It only gets called for stream disconnects.

Errors processing an individual message need their own retry schedule. We shouldn't restart the GRPC stream just because one message failed to save to the DB.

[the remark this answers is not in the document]
neekolas: I did a bit of a cleanup to at least make it more readable. There still are two layers of retry schedules. but I think that's inevitable because we do actually want two types of retries (one for the stream, one for the message)

pkg/sync/syncWorker.go
```diff
@@ -342,7 +344,14 @@ func (s *syncWorker) setupStream(
 			err,
 		)
 	}
-	originatorStream := &originatorStream{nodeID: nodeID, stream: stream}
+	expBackoff := backoff.NewExponentialBackOff()
```
mkysel, line 347: setupStream itself is wrapped in a backoff. Which itself is wrapped in an endless loop. Why do we need a third layer of retries?

pkg/sync/syncWorker.go
```diff
@@ -389,9 +398,15 @@ func (s *syncWorker) listenToStream(
 				zap.String("peer", node.HttpAddress),
 				zap.Any("numEnvelopes", len(envs.Envelopes)),
 			)
+
 			for _, env := range envs.Envelopes {
-				// ignore errors and proceed with the next message
-				_ = s.validateAndInsertEnvelope(originatorStream, env)
+				// Retry errors that may be transient up to 5X.
+				// If the error is permanent, we swallow it in the attemptEnvelopeProcessing function and allow the stream
+				// to continue. If it is transient and fails 5X, we return the error and halt the stream.
```
mkysel, line 405: it won't actually halt the stream. The top level retry will just retry again and keep failing infinitely

pkg/sync/syncWorker.go
neekolas: We should be able to retry saving an individual message without needing to go back to the other node and retrieve a fresh page of results.

pkg/sync/syncWorker.go
neekolas: Maybe halt is too strong a word. Disconnect would be better.

Returning an error here will disconnect from the other node and then it will reconnect (with a backoff) infinitely

pkg/sync/originatorStream.go
```diff
@@ -0,0 +1,307 @@
...
+					continue
+				}
+
+			storeLoop:
+				for {
+					err := s.storeEnvelope(env)
+					if err != nil {
+						s.log.Error("error storing envelope", zap.Error(err))
+						time.Sleep(1 * time.Second)
+						continue
```
mkysel, line 92: this might have to check for ctx.Done as well

mkysel (approved):
other than the shutdown I have no objections

pkg/sync/originatorStream.go
neekolas: Good catch

PR #712, opened 2025-04-11
Add pkg merkle

fbac:
### Implement Merkle tree data structure with proof generation and verification in `pkg/merkle`
* Introduces core Merkle tree implementation in [tree.go](https://github.com/xmtp/xmtpd/pull/712/files#diff-4a0ec8af27f891e4481136ec6a06b2eea6423d8992b8ea8c5c14d841b05b4d9a) with `MerkleTree` struct and tree construction logic
* Adds Keccak-256 hashing functions in [hash.go](https://github.com/xmtp/xmtpd/pull/712/files#diff-4381bdd78bb0be4de1b1e9d2b8a01153fb880ab9a8ac307d30baaa3307ccd38d) for leaf and internal node hashing
* Implements proof generation and verification in [proof.go](https://github.com/xmtp/xmtpd/pull/712/files#diff-c26afbea2e9b24a09656eaec6e2b648112cb8cf19fc1d0405b9f686e6530397d) with `MultiProof` struct
* Creates specialized proof handling for sequential elements in [proof_sequential.go](https://github.com/xmtp/xmtpd/pull/712/files#diff-f8864c26a7c602c0264a8e986cb80773e07ef20ffbf13989ee464788aecf837a)
* Adds index-based proof generation and verification in [proof_indices.go](https://github.com/xmtp/xmtpd/pull/712/files#diff-df289bc46e3f57b4c03db92365856f8c84e600ad2b9b7d8119640f2f113537c3)
* Includes comprehensive test coverage in [tree_test.go](https://github.com/xmtp/xmtpd/pull/712/files#diff-aa27c0aa647f6683e3d149d6596a81a4837849a512296c4f6419f26e088a8c7d)

#### 📍Where to Start
Start with the `NewMerkleTree` function in [tree.go](https://github.com/xmtp/xmtpd/pull/712/files#diff-4a0ec8af27f891e4481136ec6a06b2eea6423d8992b8ea8c5c14d841b05b4d9a) which implements the core tree construction logic and serves as the entry point for the package.

----

_[Macroscope](https://app.macroscope.com) summarized 5eb4a35._

## Summary by CodeRabbit

- **New Features**
	- Introduced cryptographic hashing functions with distinct prefixes for leaf and node data to support Merkle tree integrity.
	- Added a Merkle tree data structure with balanced tree construction and multi-proof generation for both sequential and arbitrary leaf indices.
	- Implemented comprehensive multi-proof verification with robust validation and error handling.
- **Tests**
	- Added extensive tests covering tree construction, hashing correctness, multi-proof generation and verification, edge cases, and utility functions to ensure reliability and correctness.


pkg/merkle/tree.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// getLeafCount returns the number of leaves in a tree.
+func getLeafCount(elementCount int) int {
+	return int(roundUpToPowerOf2(uint32(elementCount)))
+}
+
+// roundUpToPowerOf2 rounds up a number to the next power of 2.
+// Rounding up to the next power of 2 is necessary to ensure that the tree is balanced.
+func roundUpToPowerOf2(number uint32) uint32 {
```
mkysel, line 128: fancy bit smearing! 

I like this one more than the [btcd](https://github.com/btcsuite/btcd/blob/master/blockchain/merkle.go#L48) float conversion

mkysel: This might be even faster:
```
func roundUpToPowerOf2(n uint32) uint32 {
	n--
	n |= n >> 1
	n |= n >> 2
	n |= n >> 4
	n |= n >> 8
	n |= n >> 16
	n++
	return n
}
```

pkg/merkle/hash.go
```diff
@@ -0,0 +1,26 @@
+package merkle
+
+import (
+	"golang.org/x/crypto/sha3"
+)
+
+// LEAF_PREFIX is the prefix for leaf nodes.
+// It is used to distinguish leaf nodes from other types of nodes.
+const LEAF_PREFIX = "leaf|"
```
mkysel, line 9: you can't really retrieve the fact that some cryptographic hash has a `LEAF|` prefix by looking at the hash. So you can write it, but you can't ever read it out again.

What purpose does it serve?

pkg/merkle/tree.go
```diff
@@ -0,0 +1,150 @@
...
+
+import (
+	"errors"
+	"math"
+	"math/bits"
+)
+
+// MerkleTree represents a Merkle tree data structure.
+type MerkleTree struct {
+	tree      [][]byte
```
mkysel, line 11: my brain refuses to process anything related to `[][]`. 

If I were writing this, I would typedef the node so at least its `tree []Node`

Even if it is as simple as 
```
type Node struct {
	Hash []byte
}
```

pkg/merkle/hash.go
mkysel: we can get into the weeds of cryptography here, but a standalone `prefix` can not solve domain `separation`.

If we assume that the purpose of this prefix is to prevent a malicious `leaf input` from colliding with an internal node operation, it does not quite work that way.

`leaf|` aka `0x4C4541467C` is just a prefix. If you can find an internal node that starts with exactly these bytes you can craft a payload that is "(nothing instead of leaf|) || leaf1-suffix || leaf2) and you will get a collision.

I think that you need to prefix both the leaf and the internal node to achieve true domain separation.

mkysel: or just double hash :D https://github.com/OpenZeppelin/merkle-tree/blob/master/src/hashes.ts

pkg/merkle/tree.go
```diff
@@ -0,0 +1,150 @@
...
+			tree[i] = tree[leftChildIndex]
+		}
+	}
+
+	return tree, depth
+}
+
+// getDepth calculates the depth of a Merkle tree based on element count.
+func getDepth(elementCount int) int {
+	return int(math.Ceil(math.Log2(float64(elementCount))))
```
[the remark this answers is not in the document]
mkysel: @macroscopeapp what do you think?

pkg/merkle/hash.go
mkysel: Hey @macroscopeapp. Hey @coderabbitai. Compare the correctness, structural integrity and safety of Hash(LEAF|payload) == Hash(leftHash || rightHash) vs Hash(LEAF|payload) == Hash(NODE|leftHash || rightHash) vs Hash(payload) == Hash(leftHash || rightHash)

mkysel:
@coderabbitai can you generate a random stress test for the merkle tree implementation? Keep generating trees, adding nodes and asking for proofs for a configurable duration. The purpose of the stress test is to look for edge conditions and out-of-bounds errors. If you can, sanity check as much as possible.

pkg/merkle/tree_test.go
```diff
@@ -0,0 +1,331 @@
+package merkle
```
mkysel, line 1: Nit: I prefer using `merkle_test` so you can't muck with the internals of the package. There is no right or wrong way, we have just been moving a lot of tests from the the same package into its own (_test) for exactly this reason.

mkysel:
@coderabbitai here is a set of rules https://github.com/xmtp/xmtpd/blob/main/.cursor/rules/100-go-mistakes.mdc, you can also find them here https://100go.co/. Can you check this PR for any violations and if you find any, make relevant comments inline?

pkg/merkle/tree.go
```diff
@@ -0,0 +1,170 @@
...
+			tree[i] = tree[leftChildIndex]
+		}
+	}
+
+	if tree[1] == nil {
+		return nil, ErrTreeRootNil
+	}
+
+	return tree, nil
+}
```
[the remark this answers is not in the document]
fbac: Study this algorithm again and check if the scenario can exist, where rightChildIndex is not nil while leftChildIndex is nil.
Note that the algorithm fill leaf nodes from left to right, so in my opinion this won't ever happen.

pkg/merkle/proof.go
```diff
@@ -0,0 +1,238 @@
...
+	// Mark provided indices as known.
+	var (
+		leafCount = len(tree) >> 1
+		proofs    [][]byte
+		known     = make([]bool, len(tree))
+	)
+
+	for _, idx := range idxs {
+		known[leafCount+idx] = true
+	}
```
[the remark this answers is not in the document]
fbac: leafCount (len(tree) / 2) is used to access the known map (len(tree)) with leafCount+idx. Indices are previously validated to be inside bounds of the tree.
Further checks seems redundant.

pkg/merkle/proof.go
```diff
@@ -0,0 +1,313 @@
...
+func (iv indexedValues) Indices() []int {
+	indices := make([]int, len(iv))
+	for i, v := range iv {
+		indices[i] = v.index
+	}
+	return indices
+}
+
+// Verify verifies a MultiProof against the given tree root.
+func Verify(root []byte, proof *MultiProof) (bool, error) {
```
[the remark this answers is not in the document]
fbac: The check len(root) == nil also checks for nil, as a nil []byte has length 0.

pkg/merkle/tree.go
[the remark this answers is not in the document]
fbac: In the end getDepth is not needed, so I removed the code.

pkg/merkle/proof.go
```diff
@@ -0,0 +1,345 @@
...
+	indices := make([]int, count)
+	for i := 0; i < count; i++ {
+		indices[i] = startingIndex + i
+	}
+
+	return indices, nil
+}
+
+// validateIndices validates the indices slice of a proof.
+func validateIndices(indices []int, leafCount int) error {
```
neekolas, line 281: What about gaps in the list. For example, if I provide `1, 2, 4`. Those would not have duplicates, or out of bounds, but should be invalid for a sequential multi-proof.

Or is the goal to handle that validation elsewhere so that we can generate adversarial test cases?

pkg/merkle/hash.go
neekolas: Great call out

fbac: This is really interesting. :)

I wrote a new HashNode which uses a prefix for nodes, and also. uses a buffer for cheap optimizations:

```
func HashNode(left, right []byte) []byte {
	nodePrefixLen := len(nodePrefixBytes)
	buffer := make([]byte, nodePrefixLen+len(left)+len(right))

	copy(buffer[:nodePrefixLen], nodePrefixBytes)
	copy(buffer[nodePrefixLen:], left)
	copy(buffer[nodePrefixLen+len(left):], right)

	return Hash(buffer)
}
```

It settles around ~278.0 ns/op, while the old HashNode marks 280 ns/op, with better security.

pkg/merkle/proof.go
fbac: Currently we support two multi proofs; sequential and indices based. That's why we don't explicitly check the provided indices are always sequential.

We can be more opinionated and just do sequential indices.

pkg/merkle/proof.go
```diff
@@ -0,0 +1,339 @@
...
+		return ErrNoIndices
+	}
+
+	if hasDuplicates(sortedIndices) {
+		return ErrDuplicateIndices
+	}
+
+	if hasOutOfBounds(sortedIndices, balancedLeafCount) {
+		return ErrIndicesOutOfBounds
+	}
```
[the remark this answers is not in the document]
fbac: padded leaves are stil valid and should be taken in account, they're initialized as []byte{}.

pkg/merkle/proof.go
neekolas: For sequential proofs, we need to check _somewhere_ that they are actually sequential or someone could exclude a Payer from settlement. Unclear whether to me whether that should be here or elsewhere.

neekolas (approved):

pkg/merkle/tree.go
```diff
@@ -0,0 +1,320 @@
...
+func (m *MerkleTree) Root() []byte {
+	return m.tree[1].Hash()
+}
+
+// LeafCount returns the number of leaves in the Merkle tree.
+func (m *MerkleTree) LeafCount() int {
+	return len(m.leaves)
+}
+
+// BuildTree builds a serialized Merkle tree from an array of leaf nodes.
```
mkysel, line 188: seems to be called `makeTree` now

pkg/merkle/tree.go
```diff
@@ -0,0 +1,320 @@
...
+	for i := len(leaves); i < balancedLeafCount; i++ {
+		balancedLeaves[i] = []byte{}
+	}
+
+	return balancedLeaves, nil
+}
+
+// CalculateBalancedNodesCount returns the number of nodes in a balanced tree.
+// To calculate the number of nodes in a tree, we need to round up to the next power of 2.
+// Returns an error if the element count is too large to be represented in a uint32.
```
mkysel, line 284: these type casts here... It's such a shame that Go `len()` is `int` instead of `uint`. I understand the design decision, but it makes some things a bit more annoying to cast.

I would assume that all of our code will be running on 64-bit architecture and so int will be int64. `roundUpToPowerOf2` can be easily adapted to accept and return `uint64`. Just add a few more shifts :)

That might remove the need for the overflow check and a bunch of these casts.

mkysel:
```
func roundUpToPowerOf2(n int) int {
	if n <= 0 {
		return 1 // Smallest valid power of 2
	}

	// Use appropriate width based on architecture
	if bits.UintSize == 32 {
		v := uint32(n)
		if bits.OnesCount32(v) == 1 {
			return n
		}
		v |= v >> 1
		v |= v >> 2
		v |= v >> 4
		v |= v >> 8
		v |= v >> 16
		return int(v + 1)
	} else {
		v := uint64(n)
		if bits.OnesCount64(v) == 1 {
			return n
		}
		v |= v >> 1
		v |= v >> 2
		v |= v >> 4
		v |= v >> 8
		v |= v >> 16
		v |= v >> 32
		return int(v + 1)
	}
}
```

pkg/merkle/tree_test.go
mkysel, line 1: Slapping this comment here.

I would love to see a randomized test. This code is really complex and my human eyes can't see all edge cases.

mkysel:
I've been trying to figure out where a malicious actor can inject payloads. It's hard to reason about because so many functions are exposed (public). I assume you have done that because I suggested to have the tests live in their own package and test the externally exposed functionality. 

Can you explain what the entry points are? The two MultiProof functions?

We could always formalize that by using an `interface`.

pkg/merkle/tree.go
fbac: > I would assume that all of our code will be running on 64-bit architecture and so int will be int64. roundUpToPowerOf2 can be easily adapted to accept and return uint64. Just add a few more shifts :) Or make it accept int and handle both architectures.

My assumption too, extracting this to an issue to work on it.

mkysel (approved):

Issue #726, opened 2025-04-21
Merkle Tree: unify types to uint64


fbac:
these type casts here... It's such a shame that Go `len()` is `int` instead of `uint`. I understand the design decision, but it makes some things a bit more annoying to cast.

I would assume that all of our code will be running on 64-bit architecture and so int will be int64. `roundUpToPowerOf2` can be easily adapted to accept and return `uint64`. Just add a few more shifts :) Or make it accept `int` and handle both architectures.

That might remove the need for the overflow check and a bunch of these casts.

_Originally posted by @mkysel in https://github.com/xmtp/xmtpd/pull/712#discussion_r2049192872_

PR #733, opened 2025-04-22
CalculateBalanceNodesCount to handle int

fbac:
Closes https://github.com/xmtp/xmtpd/issues/726

I'd assume our code will always run in 64 bit hardware, but it's non cost to also make this package compatible with 32 bits, as we don't expect such huge trees anyway.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved input validation for node count limits, now using signed 32-bit integer boundaries for more accurate error handling.
  - Updated error messages to clearly indicate the new maximum allowed value.
- **Tests**
  - Adjusted tests to check for the updated error message and relaxed the assertion to allow for message changes.

### DIFF
--- a/pkg/merkle/tree.go
+++ b/pkg/merkle/tree.go
@@ -279,22 +279,22 @@ func makeLeaves(leaves []Leaf) ([]Leaf, error) {
 
 // CalculateBalancedNodesCount returns the number of nodes in a balanced tree.
 // To calculate the number of nodes in a tree, we need to round up to the next power of 2.
-// Returns an error if the element count is too large to be represented in a uint32.
+// Returns an error if the element count is too large to be represented in a int32.
 func CalculateBalancedNodesCount(count int) (int, error) {
 	if count <= 0 {
 		return 0, fmt.Errorf("count must be greater than 0")
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/merkle/tree.go
```diff
@@ -279,22 +279,22 @@ func makeLeaves(leaves []Leaf) ([]Leaf, error) {
...
 	}
 
-	return int(roundUpToPowerOf2(uint32(count))), nil
+	return roundUpToPowerOf2(count), nil
 }
 
 // roundUpToPowerOf2 rounds up a number to the next power of 2.
-func roundUpToPowerOf2(n uint32) uint32 {
-	if bits.OnesCount32(n) == 1 {
+func roundUpToPowerOf2(n int) int {
```
mkysel, line 296: @coderabbitai is this function correct for 64 bit numbers?

mkysel: @macroscopeapp does this work with 64 bit numbers?

mkysel: @coderabbitai try again

mkysel (approved):

PR #738, opened 2025-04-23
Add poor's man identity updates stress test

fbac:


neekolas (approved):
Have you run this enough to be sure that there aren't any nonce-related problems if some transactions revert or fail?

fbac:
> Have you run this enough to be sure that there aren't any nonce-related problems if some transactions revert or fail?

There are "nonce too high" problems, because the goroutines are executed without specific order and the mempool rejects when there's a nonce too high. 
I've observed that for tests with less than 50 parallel transaction all the tx's usually pass, it gets trickier with more than that. We could improve this by having some order, injecting more than one priv key, etc.

mkysel (approved):
LGTM. Lets take this puppy for a ride

PR #753, opened 2025-05-05
Implement 90-day message expiry for non-commit messages

mkysel:
Set expiry either provided by the payer or set it to `math.MaxInt64` meaning "forever". Identity updates and commits live forever for now. All other message types are set by the payer.

Tests will follow along with sanity validation.


mkysel:
closing for now. Will re-open once  Ihave the new approach

PR #774, opened 2025-05-09
Initial prune implementation

mkysel:
I expect to make a few changes as we progress such as dry-run and a limit to the number of cycles it takes.

## Summary by CodeRabbit

- **New Features**
  - Introduced configurable batch pruning of expired envelopes with options for maximum cycles and dry run mode.
- **Enhancements**
  - Added detailed logging of expired envelope counts and pruning progress during execution.
- **Chores**
  - Improved internal expired data management with efficient batch deletion and concurrency-safe locking.


pkg/db/sqlc/prune.sql
```diff
@@ -0,0 +1,21 @@
...
+      AND expiry < EXTRACT(EPOCH FROM now())::bigint
+    ORDER BY expiry
+    LIMIT 1000
+        FOR UPDATE SKIP LOCKED
+)
+DELETE FROM gateway_envelopes ge
+    USING to_delete td
+WHERE ge.originator_node_id = td.originator_node_id
+  AND ge.originator_sequence_id = td.originator_sequence_id
+RETURNING ge.originator_node_id, ge.originator_sequence_id, ge.expiry;
```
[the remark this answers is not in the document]
mkysel: I am aware of the missing index. I do not want to add more index bloat that will slow down the write path without seeing the performance in production. The write path is the  critical path. The deletion can be slow and nothing bad will happen as long as it can keep up in aggregate.

pkg/db/sqlc/prune.sql
```diff
@@ -0,0 +1,21 @@
+-- name: CountExpiredEnvelopes :one
+SELECT COUNT(*)::bigint AS expired_count
+FROM public.gateway_envelopes
+WHERE expiry IS NOT NULL
+  AND expiry < EXTRACT(EPOCH FROM now())::bigint;
+
+-- name: DeleteExpiredEnvelopesBatch :many
+WITH to_delete AS (
```
neekolas, line 8: If we broke this into two distinct queries we could run the read path on a replica and the deletion on the primary, to better isolate the slow query from prod traffic.

Does get a little messy with replication lag (deleted messages may still appear in subsequent queries).

Not a suggestion. Just a thought.

neekolas (approved):

pkg/db/sqlc/prune.sql
mkysel: I have not given too much thought to running an HA (or concurrent) version of this. The job is expected to be run as a cron job at some interval. The query is safe to be run concurrently but not optimized for that case.

`CountExpiredEnvelopes` might be an overkill. We will see. I want some diagnostics in the logs, but it is a fairly expensive full table scan. If it becomes too expensive, we can get rid of it.

mkysel: I have not given too much thought to running an HA (or concurrent) version of this. The job is expected to be run as a cron job at some interval. The query is safe to be run concurrently but not optimized for that case.

`CountExpiredEnvelopes` might be an overkill. We will see. I want some diagnostics in the logs, but it is a fairly expensive full table scan. If it becomes too expensive, we can get rid of it.

Issue #807, opened 2025-05-18
Proposed Rollout Plan


richardhuaaa:
Publishing this here so that we have a public point of reference.

# **Goals**

While the decentralized XMTP network is being rolled out, there will be a 'bridge' period when clients using the pre-decentralization network coexist with clients using the post-decentralization network. During this period, we have the following goals:

1. New and old clients should be able to communicate with each other.
1. Historical payloads from before the migration began should be available in the new network.
2. The cutover to smart contracts comes with complexity and risk (performance, cost). It would be great if we could control the smart contract rollout timeline separately to the client rollout timeline.

# **Concept**

1. In the new system, migrate all payloads from the old network into the new network as blockchain payloads. They have total ordering, which is the same semantics as the blockchain.
2. Until we’re ready to roll out smart contracts/incentives, we can mock reads and writes to the blockchain as reads and writes to the old XMTP network.

Due to the way sequence ID's are assigned on the old network, each payload type would use a separate sequence and be migrated as a separate originator ID. Application messages and commits are treated as the same payload type on the old node, and would share an originator ID (matching the originatorID of the ‘GroupMessages’ contract that was originally only intended for commits). This is fine. They will be ordered correctly against payloads from multiple originators using [the normal algorithm](https://community.xmtp.org/t/xip-49-decentralized-backend-for-mls-messages/856#p-2045-timestamp-ordering-55).

# **Execution**

What is the smart contract interface?

- For reads: the new node gets notified of these via an ‘indexer’.
- For writes: all client writes, smart contract or otherwise, will go through a ‘payer’.

**Milestone 1: Clients still write to the old node, but read via the new node**

Write a ‘*migration indexer*’ for the new node and swap out the existing blockchain indexer.

All it does is query for payloads from a read replica of the old node database, and return them with the same originatorID as the smart contract would use. The new node will happily write these to its DB as if it came from a smart contract. We don’t need to pay fees for this migration.

With this milestone we test the read path (and part of the write path) and clients can immediately benefit from totally ordered streams. We will always need a variation of this step no matter what because we need some way to get old payloads into the new system.

**Milestone 2: Clients write via the payer interface instead of the old node API**

On the server, modify our existing payer to make a ‘*migration payer*’:

1. If it’s an application message/key package/welcome, write it to the new node (we have this code already)
2. If it’s a blockchain payload, write it to the old node (using old node’s API) instead of to the blockchain.

Specify the migration payer as the client’s default payer.

[...]

PR #808, opened 2025-05-18
Create AppChainReader abstraction

richardhuaaa:
In order to bridge the pre-decentralization and post-decentralization XMTP networks during the migration period, the goal is to be able to mock the AppChain using the old network (reading directly from the old node's database). See [this issue](https://github.com/xmtp/xmtpd/issues/807) for more details.

Before this database-backed AppChain is created, we need to make an `AppChainReader` abstraction so that it can be freely swapped with the existing Ethereum-backed chain reader. I needed to restructure some logic to sit behind this abstraction:
- Logic for determining the smart contract address and topic to use
- Logic for parsing log events into structured data based on contract ABI's

This PR is a pure refactor, there is no functional change here.


dev/local.env
```diff
@@ -20,6 +20,7 @@ export XMTPD_SETTLEMENT_CHAIN_NODE_REGISTRY_ADDRESS="0x8d69E9834f1e4b38443C63895
 export XMTPD_SETTLEMENT_CHAIN_RATE_REGISTRY_ADDRESS="0xE71ac6dE80392495eB52FB1dCa321f5dB8f51BAE"
 
 # Application Chain Options
+export XMTPD_APP_CHAIN_USE_DATABASE_CLIENT=false
```
richardhuaaa, line 23: This is a placeholder flag that will likely change. For example, we will need the connection string for a read replica of the old node's database.

pkg/indexer/indexer.go
```diff
@@ -157,17 +160,17 @@ type builtStreamer struct {
...
 	if err != nil {
 		return nil, err
 	}
-
 	messagesTracker, err := NewBlockTracker(
 		ctx,
-		cfg.AppChain.GroupMessageBroadcasterAddress,
+		messagesAddress,
 		querier,
 	)
```
richardhuaaa, line 175: I noticed that we are fetching the latest block number as a separate call every time we read from the blockchain, and that we're also doing DB writes of the latest block number every time we write to the database. Does this fall under the existing optimization work you were tackling @fbac?

I chose to abstract at the chain reader level rather than the whole indexer, so that even when we are using the database-backed chain reader, we are still reusing as much of the existing indexer code as possible. It does mean that we potentially inherit any performance limitations, but on the flipside any performance fixes accrue to both implementations as well.

pkg/indexer/indexer.go
```diff
@@ -70,24 +65,28 @@ func (i *Indexer) StartIndexer(
...
-	client, err := blockchain.NewClient(i.ctx, cfg.AppChain.RpcURL)
+	reader, err := blockchain.NewAppChainReader(
+		i.ctx,
+		cfg.AppChain,
+	)
 	if err != nil {
 		return err
 	}
-	builder := blockchain.NewRpcLogStreamBuilder(i.ctx, client, i.log)
+	builder := blockchain.NewRpcLogStreamBuilder(i.ctx, reader, i.log)
```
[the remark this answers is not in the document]
richardhuaaa: This is existing behavior that hasn't changed with this refactor, and is a larger effort, so I will leave it for a future PR. Any resources created here are intended to be held for the lifetime of the container regardless.

pkg/blockchain/ethAppChainReader.go
```diff
@@ -0,0 +1,139 @@
...
+
+func buildIdentityUpdatesTopic() (common.Hash, error) {
+	abi, err := iu.IdentityUpdateBroadcasterMetaData.GetAbi()
+	if err != nil {
+		return common.Hash{}, err
+	}
+	return utils.GetEventTopic(abi, "IdentityUpdateCreated")
+}
+
+// Implements AppChainReader
```
fbac, line 33: The interface implementation can be forced with `var _ AppChainReader = &EthAppChainReader{}`. That way the code won't compile if the type doesn't implement the interface.

pkg/blockchain/rpcLogStreamer.go
```diff
@@ -32,31 +28,29 @@ type RpcLogStreamBuilder struct {
...
 	contractConfigs []ContractConfig
 	logger          *zap.Logger
-	ethclient       *ethclient.Client
+	reader          AppChainReader
 }
 
 func NewRpcLogStreamBuilder(
 	ctx context.Context,
-	client *ethclient.Client,
+	reader AppChainReader,
```
fbac, line 36: In my opinion the rpc log streamer shouldn't be aware of what an app chain is. In the beginning we'll have two streamers, one for an app chain and the other for a settlement chain.

There's a more modular approach in #800, where we can inject the logic here on each specific contract, or extend the IContract interface if needed.

richardhuaaa: Ah, I wasn't aware that we would also be reading from the settlement chain. In that case `AppChainReader` can simply be renamed. I wasn't intending to make the `RpcLogStreamBuilder` appchain-specific. The interface could just as well be named `ChainReader` or `ChainLogReader`, and it could just as well read from the settlement chain too, in the same way that `ethClient` would have.

The main change that the interface introduces is, instead of querying for logs by address and topic (`FilterLogs(address, topic, ...)`), we query for logs by type of event (`FilterLogs(eventType, ...)`), and handle resolving exactly where those logs should come from internally. I could put the settlement chain behind this interface as well, and call it `HybridChainLogReader` if we have a world where the settlement chain is backed by a real blockchain, and the app chain is backed by a database.

I'm on PTO this week, but can sync up again next week!

fbac:
Closing this in favor of the new migration service!

PR #811, opened 2025-05-19
Propagate ctx cancellation to IndexLogs

fbac:
- Propagate cancellation to `IndexLogs.retry()`.
- Delete duplicated chainID fields, populate it once at app_chain.go level.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved retry logic to support cancellation, allowing operations to be stopped promptly when requested.

- **Refactor**
	- Updated logging to remove the display of chain identifiers, simplifying log output for group message and identity update broadcasters.

### DIFF
--- a/pkg/indexer/common/log_handler.go
+++ b/pkg/indexer/common/log_handler.go
@@ -145,6 +145,7 @@ func IndexLogs(
 		}
 
 		err := retry(
+			ctx,
 			contract.Logger(),
 			100*time.Millisecond,
 			contract.Address().Hex(),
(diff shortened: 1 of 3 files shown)


pkg/indexer/common/log_handler.go
```diff
@@ -167,21 +168,32 @@ func IndexLogs(
...
+		default:
+			if err := fn(); err != nil {
+				logger.Error("error storing log", zap.Error(err))
+				if err.ShouldRetry() {
+					metrics.EmitIndexerRetryableStorageError(address)
+
+					select {
+					case <-ctx.Done():
+						return ctx.Err()
+					case <-time.After(sleep):
```
mkysel, line 190: @coderabbitai can you check whether time.after works in a loop? Can you show us the relevant docs.

mkysel (approved):

PR #824, opened 2025-05-19
Fix problematic tests using risky pattern

mkysel:
### Improve concurrent test error handling in blockchain nonce manager and indexer block tracker tests to fix problematic test patterns
Implements proper error propagation and collection in concurrent test scenarios:

* Modifies `TestGetNonce_ConsumeManyConcurrent` in [nonceManager_test.go](https://github.com/xmtp/xmtpd/pull/824/files#diff-6f0f25c50ea47ff1255f1c03ad4d8c6c83e79e0ae4242c5838205f05112eff58) to collect and verify goroutine errors through a dedicated error channel instead of using in-goroutine assertions
* Updates `TestConcurrentUpdates` in [block_tracker_test.go](https://github.com/xmtp/xmtpd/pull/824/files#diff-0f76a28fcca11fa536134150580023215ae855dbd3a0109c7be71c6c927a1f1f) to handle concurrent errors via a channel-based collection mechanism and adds early termination on error detection

#### 📍Where to Start
Begin with the `TestGetNonce_ConsumeManyConcurrent` function in [nonceManager_test.go](https://github.com/xmtp/xmtpd/pull/824/files#diff-6f0f25c50ea47ff1255f1c03ad4d8c6c83e79e0ae4242c5838205f05112eff58) as it contains the core pattern being fixed.

----

_[Macroscope](https://app.macroscope.com) summarized f32de70._

## Summary by CodeRabbit

- **Tests**
  - Improved error handling in concurrent test scenarios to ensure all errors are collected and reported after test completion, resulting in clearer and more synchronized test results.

### DIFF
--- a/pkg/blockchain/nonceManager_test.go
+++ b/pkg/blockchain/nonceManager_test.go
@@ -162,17 +162,28 @@ func TestGetNonce_ConsumeManyConcurrent(t *testing.T) {
 
 	var wg sync.WaitGroup
 	numClients := 20
+	errCh := make(chan error, numClients)
 
 	for i := 0; i < numClients; i++ {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
 			nonce, err := nonceManager.GetNonce(ctx)
-			require.NoError(t, err)
+			if err != nil {
+				errCh <- err
+				return
+			}
 			err = nonce.Consume()
-			require.NoError(t, err)
+			if err != nil {
+				errCh <- err
+				return
+			}
 		}()
 	}
 
 	wg.Wait()
+	close(errCh)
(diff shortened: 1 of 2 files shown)


mkysel:
This is a funny one. The crux of the issue is that calling `require or Fatal` or anything that kills the thread inside of a goroutine will terminate the goroutine. This means that the main thread sits in the `wg.Wait` forever. 

There are a few patterns on how to avoid this
- don't call `require` on the goroutine
- use t.Run for each parallel thing instead
- possibly others

All other uses of goroutine spawning is correct in the codebase.

fbac (approved):

PR #828, opened 2025-05-21
Index via websocket subscription

fbac:
### Implement WebSocket subscription support in RpcLogStreamer to enable real-time event indexing
* High Impact: Adds WebSocket subscription capabilities to `RpcLogStreamer` in [rpcLogStreamer.go](https://github.com/xmtp/xmtpd/pull/828/files#diff-7c09882fdc2ca1e5719483f613a8646adbf3750ccba69267c09948639d406ebd) with exponential backoff reconnection logic and implements dual-channel event handling in [log_handler.go](https://github.com/xmtp/xmtpd/pull/828/files#diff-a31c080b060abb8bd9dfa4d6092255d7f851497acd426d985f03ee9c70389d3c)
* Medium Impact: Introduces WebSocket URL configuration and validation in [options.go](https://github.com/xmtp/xmtpd/pull/828/files#diff-6731fb6f709392ce3e37d3b0c42074cddbce566dad2bab86af24ba7585eeb57c) and [validation.go](https://github.com/xmtp/xmtpd/pull/828/files#diff-3e0b3707cc5d712d06d1eeb6a67c10b45a7ba0b27e972924d71df3c8e539f23b)
* Low Impact: Updates test configurations and environment variables across multiple files to use WebSocket protocol instead of HTTP

#### 📍Where to Start
Start with the `RpcLogStreamer` implementation in [rpcLogStreamer.go](https://github.com/xmtp/xmtpd/pull/828/files#diff-7c09882fdc2ca1e5719483f613a8646adbf3750ccba69267c09948639d406ebd) which contains the core WebSocket subscription logic and handles both backfill and subscription channels.

----

_[Macroscope](https://app.macroscope.com) summarized a3b5dc4._

## Summary by CodeRabbit

- **New Features**
  - Added support for real-time subscription-based log streaming alongside existing backfill mechanisms, enabling near real-time event processing.
  - Introduced dedicated subscription channels for contract event handling and log indexing.
  - Added WebSocket URL fields to configuration with deprecation notices for HTTP URLs.

- **Refactor**
  - Unified event streaming into a single event channel per contract with robust subscription management including retry and backoff.
  - Updated event handling to use separate channels for event and reorg notifications with improved context cancellation and channel management.
  - Standardized naming conventions and improved logging for contract configuration fields.
  - Reorganized test setups to accommodate dual log channels and structured test state.
  - Enhanced log indexing to handle reorg events with clearer control flow and metrics emission.

- **Bug Fixes**
  - Improved validation for blockchain connection URLs by actively testing WebSocket connectivity.

- **Chores**
  - Updated environment and test configurations to use WebSocket URLs instead of HTTP for blockchain connections.
  - Declared the WebSocket package as a direct dependency.

### DIFF
--- a/pkg/blockchain/rpcLogStreamer.go
+++ b/pkg/blockchain/rpcLogStreamer.go
@@ -138,29 +138,13 @@ func (r *RpcLogStreamer) watchContract(cfg ContractConfig) {
 		logger    = r.logger.With(zap.String("contractAddress", cfg.Address.Hex()))
 		subLogger = logger.Named("subscription").
 				With(zap.String("contractAddress", cfg.Address.Hex()))
-		isSubEnabled         = false
-		backfillStartBlock   = cfg.FromBlock
-		highestBackfillBlock uint64
+		backfillFromBlock    = cfg.FromBlock
+		backfillHighestBlock = uint64(0)
+		unsafeReorgDistance  = uint64(600) // TODO: Configured per chain.
 	)
-
-	if backfillStartBlock > 0 {
-		highestBackfillBlock = backfillStartBlock - 1
-	}
 
 	defer close(cfg.eventChannel)
 	defer close(cfg.reorgChannel)
-
-	backfillEndBlock, err := r.client.BlockNumber(r.ctx)
-	if err != nil {
-		logger.Error("failed to get highest block", zap.Error(err))
-		return
-	}
-
-	innerBackfillCh, cancelBackfill, err := r.buildBackfillChannel(cfg, backfillStartBlock)
-	if err != nil {
-		logger.Error("failed to retrieve historical logs", zap.Error(err))
-		return
-	}
 
 	innerSubCh, err := r.buildSubscriptionChannel(cfg, subLogger)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/blockchain/rpcLogStreamer.go
```diff
@@ -134,83 +133,193 @@ func (r *RpcLogStreamer) Stop() {
...
-			)
 
-		case reorgBlock, open := <-watcher.reorgChannel:
+		case reorgBlock, open := <-cfg.reorgChannel:
 			if !open {
 				logger.Debug("Reorg channel closed")
 				return
 			}
-			fromBlock = reorgBlock
+			backfillStartBlock = reorgBlock
```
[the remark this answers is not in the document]
fbac: This is a redundant if. As highestBackfillBlock is defined as an uint, its default value is 0.

pkg/blockchain/rpcLogStreamer.go
```diff
@@ -167,49 +192,138 @@
...
+				zap.Uint64("fromBlock", log.BlockNumber),
 			)
-			for _, log := range logs {
-				watcher.backfillChannel <- log
+
+			cfg.subChannel <- log
+
+			// TODO: Does the timer makes sense while using a subscription with keepalive?
+			timer.Reset(cfg.MaxDisconnectTime)
+
```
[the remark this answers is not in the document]
fbac: A log buffer will be implemented in a subsequent PR. The log buffer will have at least the size of lagFromHighestBlock.

pkg/blockchain/rpcLogStreamer.go
```diff
@@ -134,83 +133,194 @@ func (r *RpcLogStreamer) Stop() {
...
-				logger.Error(
-					"Error getting next page",
-					zap.Uint64("fromBlock", fromBlock),
-					zap.Error(err),
-				)
-				time.Sleep(ERROR_SLEEP_TIME)
+
+		case log, open := <-innerSubCh:
+			if !open {
+				subLogger.Fatal("subscription channel closed, closing watcher")
```
mkysel, line 192: does fatal call panic?

This also has to close all channels. Or the better idea might be to call defer on the cancel functions after creation.

pkg/blockchain/rpcLogStreamer.go
```diff
@@ -134,83 +133,194 @@ func (r *RpcLogStreamer) Stop() {
...
-				"Got logs",
-				zap.Int("numLogs", len(logs)),
-				zap.Uint64("fromBlock", fromBlock),
-				zap.Time("time", time.Now()),
-			)
-			for _, log := range logs {
-				watcher.backfillChannel <- log
+			if log.BlockNumber > highestBackfillBlock {
+				highestBackfillBlock = log.BlockNumber
+			}
```
mkysel, line 230: can the backfill be out of order? Why would log.Blocknumber ever be lower than the one before?

pkg/blockchain/rpcLogStreamer.go
```diff
@@ -134,83 +133,194 @@ func (r *RpcLogStreamer) Stop() {
 	r.wg.Wait()
 }
 
-func (r *RpcLogStreamer) watchContract(watcher ContractConfig) {
+func (r *RpcLogStreamer) watchContract(cfg ContractConfig) {
```
mkysel, line 136: I think that I can break this algorithm.

client.BlockNumber and channel creation are not atomic.

- Lets say that backfillEndBlock gives us block 100
- 10 blocks get inserted concurrently
- We construct backfill at 90, want to go to 100
- Subscription constructor takes a long time to start. It would start at 110, but it doesnt give us that data in a channel yet
- Innerbackfillch gives us 90-100 as designed
- Innerbackfill reaches highestBackfillBlock(100) >= backfillEndBlock(100)
- We enable sub
- Sub finally opens a channel and starts at 110

mkysel: the order of operations has to be:
1) open the sub, guarantee via a waitgroup that the channel has been opened
2) get the current highest block in case there are no new events coming from the sub
3) open backfill

mkysel:
```
func (r *RpcLogStreamer) watchContract(cfg ContractConfig) {
	var (
		logger    = r.logger.With(zap.String("contractAddress", cfg.Address.Hex()))
		subLogger = logger.Named("subscription").
				With(zap.String("contractAddress", cfg.Address.Hex()))
		backfillStartBlock = cfg.FromBlock
	)

	defer close(cfg.eventChannel)
	defer close(cfg.reorgChannel)

	innerSubCh, err := r.buildSubscriptionChannel(cfg, subLogger)
	if err != nil {
		logger.Error("failed to subscribe to contract", zap.Error(err))
		return
	}

	// backfillEndBlock is inclusive from backfill
	backfillEndBlock, err := r.client.BlockNumber(r.ctx)
	if err != nil {
		logger.Error("failed to get highest block", zap.Error(err))
		return
	}

	innerBackfillCh, cancelBackfill, err := r.buildBackfillChannel(cfg, backfillStartBlock)
	if err != nil {
		logger.Error("failed to retrieve historical logs", zap.Error(err))
		return
	}

	logger.Info("Starting watcher")

	// 1) fully consume the backfill until we reach the best known state after the sub was created
watchLoop:
	for {
		select {
		case <-r.ctx.Done():
			logger.Debug("Context cancelled, stopping watcher")
			return

		case _, _ = <-cfg.reorgChannel:
			//TODO
			return

		case log, open := <-innerBackfillCh:
			if !open {
				logger.Fatal("backfill channel closed, closing watcher")
				return
			}

			cfg.eventChannel <- log

			// Check if subscription has to be enabled only after processing all logs in the last batch.
			// Duplicated logs are not a problem, lost logs are.
			if log.BlockNumber >= backfillEndBlock {
				logger.Debug(
					"Backfill complete, enabling subscription mode",
					zap.Uint64("blockNumber", log.BlockNumber),
				)

				cancelBackfill()
				break watchLoop
			}
		}
	}

	// 2) now we can safely use the sub
	for {
		select {
		case <-r.ctx.Done():
			logger.Debug("Context cancelled, stopping watcher")
			return

		case _, _ = <-cfg.reorgChannel:
			//TODO
			return

		case log, open := <-innerSubCh:
			if !open {
				subLogger.Fatal("subscription channel closed, closing watcher")
				return
			}

			
			// already processed by backfill
			if log.BlockNumber <= backfillEndBlock {
				subLogger.Debug(
					"Log already processed",
					zap.Uint64("fromBlock", log.BlockNumber),
				)
				continue
			}

			subLogger.Debug(
				"Sending log to subscription channel",
				zap.Uint64("fromBlock", log.BlockNumber),
			)
			
			cfg.eventChannel <- log
		}
	}
}
```

^^^ it does not account for reorgs, but is much more readable to me

pkg/blockchain/rpcLogStreamer.go
fbac: Good catch as well, switching to logger.Error + return.

pkg/blockchain/rpcLogStreamer.go
fbac: Initially I had a similar approach, but discarded it. I don't think it would work, specially during long backfills.

As I see it: 
- We set `backfillEndBlock` once, which is the latest block at the time r.client.BlockNumber() is fired. 
- It would be never updated afterwards, the process backfills from start to end
- In the meanwhile the blockchain will keep producing blocks, so in practice the real `backfillEndBlock` has increased, and we'll miss those logs. There is no way as well to signal the subscription to keep those logs in a buffer.

By the time we switch to the subscription channel it would be advanced and we'd be receiving logs from the initial `backfillEndBlock` + N blocks ahead. 
N can be huge if the backfill process took long enough (and it will), which is possible for new nodes syncing 1 year historical data (126,230,400 blocks at 250ms per block, worst case scenario)

That's the whole reason around moving the backfill and sub channel to the same select, and having `backfillEndBlock` as a moving target thanks to the logs coming to the sub channel.

This gave me an idea though, which might be better than the current approach and combines both. 
Theoretically we could "weight" the channels, like:

```
backfillLoop:
for {
    // Subscription channel updates backfillEndBlock to the latest log seen.
    select {
    case log := <-subChan:
        // Update backfillEndBlock.
    default:
        // There was no log in subChan, backfill.
    }

    // Next backfill batch.
    select {
    case log := <-backfillChan:
        // Process logs.
        // if log.BlockNumber >= backfillEndBlock
        // cancel backfill and break backfillLoop
    }
}

// Switch to a subscription-only.
for {
    select {
    case log := <-subChan:
        // A backfill log arrived after we found none initially
        processBackfill(log)
    }
}
```

pkg/blockchain/rpcLogStreamer.go
```diff
@@ -134,83 +133,194 @@ func (r *RpcLogStreamer) Stop() {
...
-				watcher.backfillChannel <- log
+			if log.BlockNumber > highestBackfillBlock {
+				highestBackfillBlock = log.BlockNumber
+			}
+
+			cfg.eventChannel <- log
+
+			// Check if subscription has to be enabled only after processing all logs in the last batch.
+			// Duplicated logs are not a problem, lost logs are.
+			if highestBackfillBlock+uint64(r.lagFromHighestBlock) >= backfillEndBlock {
```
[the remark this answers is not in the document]
fbac: You're right, and after some testing I believe the problem is deeper.

We currently backfill with the configurable lag, in GetNextPage:
```
highestBlockCanProcess := highestBlock - uint64(r.lagFromHighestBlock)
if fromBlock > highestBlockCanProcess {
	metrics.EmitIndexerCurrentBlockLag(contractAddress, 0)
	return []types.Log{}, nil, nil
}
```

This isn't compatible with a websocket subscription, which will always get the latest canonical and reorged events (field Removed = true|false). It's also not compatible with the app chain, where we always want 0 lag (it's hardcoded)

So I believe we should backfill always with 0 lag for every chain. 
Then, when switch over to a sub, after receiving an event in a "lagged" chain we can:
- maintain it in an icebox and send it over the channel after the safe lag has passed (lag_blocks * time_per_block)
- if before that the indexer receives that log was reorged (the same event, with Removed = true), we cancel it.

pkg/blockchain/rpcLogStreamer.go
```diff
@@ -134,116 +133,243 @@ func (r *RpcLogStreamer) Stop() {
...
+		return nil, err
+	}
+
+	tracing.GoPanicWrap(
+		r.ctx,
+		&r.wg,
+		fmt.Sprintf("rpcLogStreamer-watcher-subscription-%v", cfg.Address),
+		func(ctx context.Context) {
+			defer close(innerSubCh)
+			defer sub.Unsubscribe()
```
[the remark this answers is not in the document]
fbac: When the sub is recreated the sub var gets replaced, the deferred Unsubscribe would still apply.

mkysel (approved):

PR #866, opened 2025-06-03
Refactor large options configs using the functional options pattern

mkysel:


fbac (approved):
To me this looks correct, the options pattern is well implemented.

A nitpick however, every time I've used this pattern it was for optional settings that can be defaulted to a sane value, not using the options to inject non-optional dependencies. It's not a deal breaker though.

PR #878, opened 2025-06-06
Estimate gas before sending tx

fbac:
### Estimate gas before sending transactions by adding gas estimation and balance validation to ExecuteTransaction function in blockchain client
The `ExecuteTransaction` function in [pkg/blockchain/client.go](https://github.com/xmtp/xmtpd/pull/878/files#diff-10bd63d42f0476b000d46c9ab0629a13fcd0d38283206ae0202a52992073d990) now performs gas estimation and balance validation before sending transactions. The function adds a dry-run transaction simulation with `NoSend=true`, explicit gas estimation using `ethclient.EstimateGas`, fallback gas price determination, and validation that the sender has sufficient balance to cover gas costs. The polling interval for transaction confirmation increases from 2 seconds to 10 seconds in the `WaitForTransaction` function.

#### 📍Where to Start
Start with the `ExecuteTransaction` function in [pkg/blockchain/client.go](https://github.com/xmtp/xmtpd/pull/878/files#diff-10bd63d42f0476b000d46c9ab0629a13fcd0d38283206ae0202a52992073d990).

----

_[Macroscope](https://app.macroscope.com) summarized fa17fa6._

### DIFF
--- a/pkg/blockchain/client.go
+++ b/pkg/blockchain/client.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/xmtp/xmtpd/pkg/metrics"
 
+	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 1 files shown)


pkg/blockchain/client.go
```diff
@@ -53,27 +56,81 @@ func ExecuteTransaction(
...
+		return fmt.Errorf("failed to send tx: %w", err)
 	}
 
+	// Step 7: Wait for receipt.
 	receipt, err := WaitForTransaction(
 		ctx,
 		logger,
 		client,
-		2*time.Second,
+		10*time.Second,
```
mkysel, line 125: Yeah. Base is much slower. Should we make it configurable with a 10s default?

mkysel (approved):

PR #897, opened 2025-06-18
Add migrator service

fbac:
### Add migrator service to transfer MLS V1 data from source database to xmtpd database format with blockchain publishing support
Introduces a complete data migration service that reads data from a source database, transforms it into xmtpd envelope format, and writes it to either the destination database or blockchain. The implementation includes:

- Core migration service in [pkg/migrator/migrator.go](https://github.com/xmtp/xmtpd/pull/897/files#diff-ccb0d31b1e7491838a2e4c7f2f6523eed971d6f18d4c04f006e14e2767671693) with concurrent processing pipelines for group messages, inbox logs, welcome messages, and key packages
- Database readers in [pkg/migrator/reader.go](https://github.com/xmtp/xmtpd/pull/897/files#diff-5afb69a12b4d409ea45341dca6b796f48dc578ab56ad078142152b5628bf0828) for fetching batches of records from source tables
- Data transformers in [pkg/migrator/transformer.go](https://github.com/xmtp/xmtpd/pull/897/files#diff-c76d70079716ba42e22c5b5031c1960504cccbca874aadb118385ca67341cf62) that convert source data to XMTPD envelope format with proper signing
- Database and blockchain writers in [pkg/migrator/writer.go](https://github.com/xmtp/xmtpd/pull/897/files#diff-79de0174d18bbb97eb4c62be1eeec1e2132ff3060f5beafb3b8721044791b258) for persisting migrated data
- Migration progress tracking via new `migration_tracker` table and associated queries
- Configuration options in [pkg/config/migration.go](https://github.com/xmtp/xmtpd/pull/897/files#diff-87c3e997fe7ecbf4c63d056ef11ed04d37b5ba302027c9efd6432f1a8c15880a) for enabling and configuring the migration service
- Integration with the main server in [pkg/server/server.go](https://github.com/xmtp/xmtpd/pull/897/files#diff-104a11712684e520f0affa95e6a053faeb713306a10f69ee99b9729cc9dd1996) to start and stop the migration service
- Comprehensive test suite with database and blockchain test infrastructure

#### 📍Where to Start
Start with the `NewMigrationService` constructor function in [pkg/migrator/migrator.go](https://github.com/xmtp/xmtpd/pull/897/files#diff-ccb0d31b1e7491838a2e4c7f2f6523eed971d6f18d4c04f006e14e2767671693) to understand how the migration service is initialized, then review the `Start` method to see how the migration process begins.

----

_[Macroscope](https://app.macroscope.com) summarized 5e37275._


pkg/db_migrator/migrator.go
```diff
@@ -0,0 +1,335 @@
...
+
+						case recvChan <- record:
+						}
+					}
+
+					logger.Debug("batch sent to transformer",
+						zap.Int("total_fetched", len(records)),
+						zap.Int64("new_last_id", newLastID))
+
+					if err := dstQueries.UpdateMigrationProgress(ctx, queries.UpdateMigrationProgressParams{
```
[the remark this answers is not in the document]
fbac: Most probably this logic will be moved to the writer, specifically when the data is inserted into the destination db we'll also update the progress.

pkg/db_migrator/migrator.go
```diff
@@ -0,0 +1,335 @@
...
+func (s *dbMigrator) Start(ctx context.Context) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if s.running {
+		return fmt.Errorf("migration service is already running")
+	}
+
+	s.running = true
+
```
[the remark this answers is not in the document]
fbac: All the methods will use s.ctx and s.cancel. I removed the context passed as parameter.

pkg/migrator/transformer.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	protoClientEnvelope := &proto.ClientEnvelope{
+		Payload: &proto.ClientEnvelope_WelcomeMessage{
+			WelcomeMessage: &mlsv1.WelcomeMessageInput{
+				Version: &mlsv1.WelcomeMessageInput_V1_{
+					V1: &mlsv1.WelcomeMessageInput_V1{
+						InstallationKey: welcomeMessage.InstallationKey,
+						Data:            welcomeMessage.Data,
+						HpkePublicKey:   welcomeMessage.HpkePublicKey,
```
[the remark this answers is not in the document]
neekolas: I believe we are adding a few more fields in the next week or two as well, unfortunately.

pkg/migrator/transformer.go
```diff
@@ -0,0 +1,341 @@
...
+
+	return t.originatorEnvelope(
+		protoClientEnvelope,
+		groupMessageOriginatorID,
+		uint64(groupMessage.ID),
+	)
+}
+
+// TransformInboxLog converts InboxLog to appropriate XMTPD IdentityUpdate envelope format.
+func (t *transformer) TransformInboxLog(
```
neekolas, line 120: When we process an identity update from the blockchain (in the old node and new) we update the address log as well when addresses are added and removed from an inbox as part of an update. 

We will need to do that somewhere in this process. We could pull from the address_log table separately, but it might become difficult to reconcile with identity updates coming in from the chain. Especially since rows in the address log are mutable.

The alternative is to do the same thing we do when pulling from the chain, and get the list of added/revoked addresses by processing the update through the MLS validation service. Adds an external service call, but is closer to how this process works everywhere else.

pkg/migrator/transformer.go
```diff
@@ -0,0 +1,341 @@
...
+
+	return t.originatorEnvelope(
+		protoClientEnvelope,
+		inboxLogOriginatorID,
+		uint64(inboxLog.SequenceID),
+	)
+}
+
+// TransformInstallation converts Installation to appropriate XMTPD KeyPackage envelope format.
+func (t *transformer) TransformInstallation(
```
neekolas, line 159: In V3, when you upload a new key package we mutate the existing installation and change the key package field. That is going to be hard to make work in an ETL world. I would suggest we actually update V3 to have an append-only table of key packages joined to installations, which will be easier to migrate.

pkg/migrator/transformer.go
fbac: My idea is to inject the MLS validation service in the writer and doing it there. 
Same as in the indexer; when processing an IdentityUpdate, get the association states and act based on the diffs.

The difference is that in the indexer we get the gateway envelopes with originatorID 1 (IU from chain), and in the migrator writer we'd get the originatorID 12 since AFAIK these identity updates would only be relevant to the ones being migrated.

pkg/migrator/transformer.go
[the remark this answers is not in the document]
fbac: It was weird, spotted why this is missing and shipped a fix.

pkg/migrator/transformer.go
fbac: Makes complete sense to me. New issue created to track the work. https://github.com/xmtp/xmtpd/issues/930

pkg/migrator/transformer.go
```diff
@@ -0,0 +1,347 @@
...
+		return nil, fmt.Errorf("failed to sign client envelope: %w", err)
+	}
+
+	// TODO: Change expiration for group messages and identity updates.
+
+	retentionDays := uint32(math.MaxUint32)
+
+	if !clientEnvelope.Aad().IsCommit {
+		retentionDays = constants.DEFAULT_STORAGE_DURATION_DAYS
+	}
```
mkysel, line 285: I fixed this somewhere, so you will have to copy the new code

pkg/migrator/transformer.go
```diff
@@ -0,0 +1,347 @@
...
+
+	if groupMessage.GroupID == nil {
+		return nil, fmt.Errorf("groupID is nil")
+	}
+
+	if len(groupMessage.Data) <= 0 {
+		return nil, fmt.Errorf("data is empty")
+	}
+
+	protoClientEnvelope := &proto.ClientEnvelope{
```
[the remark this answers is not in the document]
fbac: Yes, the expiration is set in the `buildAndSignPayerEnvelope` func.

pkg/migrator/testdata/schema.sql
```diff
@@ -0,0 +1,30 @@
...
+	installation_id BYTEA NOT NULL,
+	key_package BYTEA NOT NULL
+);
+
+CREATE TABLE group_messages(
+	id BIGSERIAL PRIMARY KEY,
+	created_at TIMESTAMP NOT NULL DEFAULT NOW(),
+	group_id BYTEA NOT NULL,
+	data BYTEA NOT NULL,
+	group_id_data_hash BYTEA NOT NULL
```
neekolas, line 12: Martin now has the `is_commit` field in this DB, and I have a [PR open](https://github.com/xmtp/xmtp-node-go/pull/442) to add a new `should_push` and `sender_hmac` coming very soon.

Sorry to make you have to hit a moving target. Should be the end of schema updates soon.

pkg/migrator/transformer.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	unsignedEnv := proto.UnsignedOriginatorEnvelope{
+		OriginatorNodeId:         payerEnvelope.TargetOriginator,
+		OriginatorSequenceId:     sequenceID,
+		OriginatorNs:             time.Now().UnixNano(),
+		PayerEnvelopeBytes:       payerEnvelopeBytes,
+		BaseFeePicodollars:       0,
+		CongestionFeePicodollars: 0,
+		ExpiryUnixtime: uint64(
```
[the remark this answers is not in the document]
fbac: Check the code again, I think this is non an issue.
Identity updates and group messages which are commits are stored in the blockchain directly, the expiry unix time is MaxInt32 but is never used. They're directly stored onchain.
The rest of messages get a default of 60 days.

pkg/migrator/writer.go
```diff
@@ -0,0 +1,254 @@
...
+	originatorEnvelopeBytes, err := proto.Marshal(env.Proto())
+	if err != nil {
+		m.log.Error("marshall originator envelope failed", zap.Error(err))
+		return re.NewNonRecoverableError("marshall originator envelope failed", err)
+	}
+
+	err = db.RunInTx(
+		m.ctx,
+		m.writer,
+		nil,
```
fbac, line 53: @mkysel @neekolas should we force some type of db isolation?

PR #905, opened 2025-06-23
Close timing gap in socket discovery

mkysel:
Some of our tests fail because there is a timing gap between our test discovering a socket, and the test actually using the socket. This refactor discovers a socket and keeps the listener open until it is time to properly close it.


neekolas (approved):
Nice find

PR #911, opened 2025-06-25
Add defensive checks for valid expiration

mkysel:


pkg/api/message/service.go
```diff
@@ -369,6 +369,20 @@ func (s *Service) PublishPayerEnvelopes(
...
+		)
+	}
+
+	if topicKind == topic.TOPIC_KIND_GROUP_MESSAGES_V1 && payerEnv.ClientEnvelope.Aad().IsCommit {
+		return nil, status.Errorf(
+			codes.InvalidArgument,
+			"commit messages must be published via the blockchain",
+		)
+	}
+
```
mkysel, line 385: we should really never accept these...

mkysel: is it the same thing?

PR #923, opened 2025-06-27
Add pprof http server

fbac:


pkg/debug/pprof.go
```diff
@@ -0,0 +1,55 @@
...
+
+	select {
+	case <-ctx.Done():
+		return s.Shutdown(context.Background())
+	case err := <-errChan:
+		return err
+	}
+}
+
+func (s *Server) Shutdown(ctx context.Context) error {
```
mkysel, line 49: I don't see anything calling this

mkysel (approved):

PR #932, opened 2025-06-30
Separate DB worker from sync

mkysel:
The DB goroutine lifecycle was tied to the lifecycle of the GRPC connection. There were no guarantees that the DB conn won't outlive the main routine when being rebuilt. This could lead to concurrent DB writing coroutines, out-of-order writes and other nastiness.

### DIFF
--- a/pkg/sync/originatorStream.go
+++ b/pkg/sync/originatorStream.go
@@ -2,75 +2,54 @@ package sync
 
 import (
 	"context"
-	"database/sql"
 	"errors"
 	"io"
 	"strings"
-	"time"
 
 	"github.com/cenkalti/backoff/v5"
-	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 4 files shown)


pkg/sync/originatorStream.go
```diff
@@ -2,75 +2,54 @@ package sync
...
+		stream:       stream,
+		writeQueue:   writeQueue,
 	}
 }
 
 func (s *originatorStream) listen() error {
 	recvChan := make(chan *message_api.SubscribeEnvelopesResponse)
-	errChan := make(chan error)
-	writeQueue := make(chan *envUtils.OriginatorEnvelope, 10)
+	errChan := make(chan error, 1)
```
mkysel, line 50: guarantees that the reader routine can exit if the main routine is doing something else. Otherwise the write to err might block forever.

fbac (approved):
LGTM. Love the separation of concerns.

PR #933, opened 2025-06-30
Prevent write-to-closed-channel panics

mkysel:
First the easy stuff:
- got rid of the FixedRegistry

Now the hard stuff:
- removed the ability to cancel a sub. It is extremely difficult to reason about channels being closed by the reader. It is an anti-pattern that can lead to panics. 
- stop rebuilding the sub on every rebuild
- made the notifier use a buffered channel of 1 which removes the need for concurrent go routines.


pkg/sync/nodeRegistryWatcher.go
```diff
@@ -0,0 +1,79 @@
...
+func (w *NodeRegistryWatcher) triggerCancel() {
+	if w.cancelFn != nil {
+		w.fnLock.Lock()
+		defer w.fnLock.Unlock()
+		w.cancelFn()
+		w.cancelFn = nil
+	}
+}
+
+func (w *NodeRegistryWatcher) Watch() {
```
neekolas, line 50: I think it would be helpful to include some comments explaining how to properly use this function. Not the most intuitive to figure out when to call `watch` and what the cancel function signifies.

neekolas (approved):

fbac (approved):

PR #944, opened 2025-07-07
Add keepalive config to chain client

fbac:


fbac:
> Can you give me the context? We are disabling it explicitly and enabling by default? Is this related to the restarts we see?

By default it's enabled, I'm only disabling these in short lived connections such as the cli retrieving some data.

In xmtpd it will use the default dialer, and in the indexer those values could be adjusted to the chain block times, etc.

And yep, this is relevant to the restarts we see. I'm not confident this is 100% the solution as I think these are related to the server (blockchain) mostly. In a followup PR I'm also including the backfill on sub recreation.

PR #987, opened 2025-07-22
Validation free isCommit determination

mkysel:


mkysel:
Once this and the 16b group is merged, I will change the payer to use the same code for routing

PR #993, opened 2025-07-28
Blockchain originator IDs

fbac:
- Get the proper blockchain originator ID using constants instead of hardcoding values.
- Minor variable naming to make the formatter happy.


mkysel (approved):
no objections other than it will merge conflict with #992

PR #998, opened 2025-07-29
Teach blockchain publisher to bootstrap messages

fbac:
### Add batch publishing methods `BootstrapGroupMessages` and `BootstrapIdentityUpdates` to `BlockchainPublisher` for publishing multiple messages in single transactions
- Adds `BootstrapGroupMessages` and `BootstrapIdentityUpdates` methods to [`BlockchainPublisher`](https://github.com/xmtp/xmtpd/pull/998/files#diff-774607218d14c81b502c9f8090b1dfccd17f552b0c9bbd9d475de4b674d0a80b) enabling batch publishing of multiple group messages or identity updates in single blockchain transactions
- Replaces `findLog` function with `findLogs` function that returns multiple logs and validates expected event counts with new `ErrNoLogsFound` error handling
- Modifies existing `PublishGroupMessage` and `PublishIdentityUpdate` methods to use the new `findLogs` function and updates callback return types in `withNonce` to support slices
- Adds comprehensive test coverage for new batch operations in [`blockchainPublisher_test.go`](https://github.com/xmtp/xmtpd/pull/998/files#diff-3cddc1f859e108c4a81bc49ba7891f25a4c59d0865fb432b4a93a51c9ec22d38) with error handling validation for various scenarios

#### 📍Where to Start
Start with the new `BootstrapGroupMessages` and `BootstrapIdentityUpdates` methods in [`blockchainPublisher.go`](https://github.com/xmtp/xmtpd/pull/998/files#diff-774607218d14c81b502c9f8090b1dfccd17f552b0c9bbd9d475de4b674d0a80b) to understand the batch publishing functionality.

#### Changes since #998 opened

- Relocated log count validation from `findLogs` function to individual publisher methods [a481e8f]
- Added capacity hint to slice initialization in `findLogs` function [a481e8f]
----

_[Macroscope](https://app.macroscope.com) summarized a481e8f._

### DIFF
--- a/pkg/blockchain/blockchainPublisher.go
+++ b/pkg/blockchain/blockchainPublisher.go
@@ -168,6 +168,10 @@ func (m *BlockchainPublisher) PublishGroupMessage(
 		return nil, err
 	}
 
+	if len(logs) != 1 {
+		return nil, ErrNoLogsFound
+	}
+
 	return logs[0], nil
 }
 
@@ -273,6 +277,10 @@ func (m *BlockchainPublisher) PublishIdentityUpdate(
 		return nil, err
 	}
 
+	if len(logs) != 1 {
+		return nil, ErrNoLogsFound
+	}
+
 	return logs[0], nil
 }
 
@@ -336,7 +344,7 @@ func findLogs[T any](
 	parseFunc func(types.Log) (*T, error),
 	expectedEventCount int,
 ) ([]*T, error) {
-	events := make([]*T, 0)
+	events := make([]*T, 0, expectedEventCount)
 
 	for _, logEntry := range receipt.Logs {
 		if logEntry == nil {
@@ -353,14 +361,6 @@ func findLogs[T any](
 
 	if len(events) == 0 {
 		return nil, ErrNoLogsFound
-	}
-
-	if len(events) != expectedEventCount {
-		return nil, fmt.Errorf(
-			"expected %d logs, got %d",
-			expectedEventCount,
-			len(events),
-		)
 	}
 
 	return events, nil


pkg/blockchain/blockchainPublisher.go
```diff
@@ -155,25 +157,83 @@ func (m *BlockchainPublisher) PublishGroupMessage(
...
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return logs[0], nil
+}
+
+func (m *BlockchainPublisher) BootstrapGroupMessages(
```
neekolas, line 174: With multiple messages it becomes easier for us to exceed the limits on a single tx. Are we planning some mechanism to break up multiple payloads into smaller chunks if we get errors there?

fbac: I plan to use this function to bootstrap one message at a time. Initially I don't think we need to batch anything and risk reverting tx's.

The amount of messages to backfill can be easily migrated in a few days, and then the migrator would continue pulling these on the fly.

If down the road we're confident this can be used with batches then I'll modify it to guarantee the data published in the batch is not bigger than the max data a tx can hold.

pkg/blockchain/blockchainPublisher.go
```diff
@@ -202,40 +322,56 @@ func (m *BlockchainPublisher) PublishIdentityUpdate(
...
-func findLog[T any](
+func findLogs[T any](
 	receipt *types.Receipt,
-	parse func(types.Log) (*T, error),
-	errorMsg string,
-) (*T, error) {
+	parseFunc func(types.Log) (*T, error),
+	expectedEventCount int,
+) ([]*T, error) {
+	events := make([]*T, 0)
```
mkysel, line 339: pre-allocate the right size here?

pkg/blockchain/blockchainPublisher.go
neekolas: Glad I asked. That makes sense

Issue #913, opened 2025-06-25
Test failure: TestCanGenerateAndAttestReport is flaky


mkysel:
```
2025-06-24T21:47:48.785-0700    DEBUG   NodeRegistryAdmin       blockchain/client.go:53 Sender balance  {"address": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266", "balance": "9999988341398080099069"}
2025-06-24T21:47:48.809-0700    DEBUG   NodeRegistryAdmin       blockchain/client.go:94 Gas estimation  {"address": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266", "gas": 223877}
2025-06-24T21:47:49.074-0700    INFO    NodeRegistryAdmin       blockchain/registryAdmin.go:116 node added to registry  {"node_id": 100, "owner": "0x34C3cb6e415d1E00E545f2FbBF91E19D5050B3d2", "http_address": "http://localhost:62921", "signing_key_pub": "0481d6d745d0502f1a00990d98120add19a76c8c58f5ecc8dc462c59a9720d9a0ea9491d9114b96e76c2d5dc838d34b7b7f75876ec04040527481257f1e27d36c7"}
2025-06-24T21:47:49.076-0700    DEBUG   NodeRegistryAdmin       blockchain/client.go:53 Sender balance  {"address": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266", "balance": "9999988072933920614469"}
2025-06-24T21:47:49.088-0700    DEBUG   NodeRegistryAdmin       blockchain/client.go:94 Gas estimation  {"address": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266", "gas": 40761}
2025-06-24T21:47:49.346-0700    INFO    NodeRegistryAdmin       blockchain/registryAdmin.go:158 node added to canonical network {"node_id": 100}
2025-06-24T21:47:49.348-0700    DEBUG   NodeRegistryAdmin       blockchain/client.go:53 Sender balance  {"address": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266", "balance": "9999988019792539787739"}
2025-06-24T21:47:49.364-0700    DEBUG   NodeRegistryAdmin       blockchain/client.go:94 Gas estimation  {"address": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266", "gas": 223889}
[...]

PR #1060, opened 2025-08-22
Enable TestCanGenerateAndAttestReport

fbac:
Close https://github.com/xmtp/xmtpd/issues/913

I fixed this at some point, this was related to the way the PayerReport calculates its ID.


mkysel (approved):
we will see whether it is really fixed

PR #1074, opened 2025-08-28
Require database exists if migration is enabled

fbac:
### Require the database to exist when the migration server is enabled by triggering namespace derivation and namespaced DB initialization in `main.main`
Update `main.main` to broaden the conditional that gates namespace derivation and database initialization to also include `options.MigrationServer.Enable` alongside the existing Replication, Sync, and Indexer checks in [main.go](https://github.com/xmtp/xmtpd/pull/1074/files#diff-0d69b50851c3a886b12eba860f887169c16cf724884bd031a7be4a7f556aee85).

#### 📍Where to Start
Start at the conditional in `main.main` in [main.go](https://github.com/xmtp/xmtpd/pull/1074/files#diff-0d69b50851c3a886b12eba860f887169c16cf724884bd031a7be4a7f556aee85) that gates namespace derivation and namespaced database initialization.

#### Changes since #1074 opened

- Added required field validation for signer private key when migration server is enabled [2b93b22]
----

_[Macroscope](https://app.macroscope.com) summarized 2b93b22._

### DIFF
--- a/pkg/config/validation.go
+++ b/pkg/config/validation.go
@@ -119,6 +119,11 @@ func validateMigrationOptions(
 			missingSet,
 		)
 		validateField(
+			opts.Signer.PrivateKey,
+			"signer.private-key",
+			missingSet,
+		)
+		validateField(
 			opts.MigrationServer.ReaderConnectionString,
 			"migration-server.reader-connection-string",
 			missingSet,


cmd/replication/main.go
```diff
@@ -97,7 +97,8 @@ func main() {
 			}()
 		}
 
-		if options.Replication.Enable || options.Sync.Enable || options.Indexer.Enable {
+		if options.Replication.Enable || options.Sync.Enable || options.Indexer.Enable ||
```
mkysel, line 100: this means that the migrator needs to have the Signer.PrivateKey of the node. Otherwise we will compute a different namespace. Do we want to pass this info around?

fbac: It's needed, as the migrator has to sign envelopes PayerEnvelope and OriginatorEnvelope both.

mkysel: do we expect to have a separate payer key for this? Or are we reusing the node key?

fbac: Separate payer key, already set up in testnet. That way we can properly identify the migrator "payer" if needed.

mkysel: the signer.private key should be checked in `validateMigrationOptions`

fbac: Yes, that's done. Maybe I explained myself wrong.
The migrator has two keys:
- Separate payer key, to properly identify the migrator as an unique payer.
- Node private key, which currently is the same as node 100. It's the same key where the payer runs, and the infra reflects this.

mkysel: makes sense. Can you add these keys to `validateMigrationOptions`? Approving otherwise

mkysel (approved):

PR #1172, opened 2025-09-24
refactor: use WaitGroup.Go to simplify code

juejinyuxitu:
use WaitGroup.Go to simplify code.

More info can see: https://github.com/golang/go/issues/63796


mkysel (approved):
thanks for the simplification!

PR #1174, opened 2025-09-24
Enable workers on app startup

neekolas:


mkysel (approved):
I will make a follow-up with a few changes to tests

PR #1221, opened 2025-10-07
Fix integration test which races with payer reports

mkysel:


neekolas (approved):
That's a fun case to worry about. Maybe we want to disable the payer report generator for at least some of these tests (the ones truly unrelated) just to reduce chaos

PR #1248, opened 2025-10-15
Allow configuring environment from baked in options

neekolas:
## tl;dr

- Allows configuration using pre-baked environments, where the config will be included in the docker image. You can now specify `--contracts.environment=testnet-staging` and get the latest config.
- This will allow users to get the latest chain config simply by upgrading to the latest SDK, which should make node operator and gateway operator's lives easier
- I hope this will also help us keep the ABIs and the contracts we are talking to in sync easier, since you can't use this field with an older contract address (maybe the proxies negate some of this benefit).

## Design considerations

- I didn't want to remove any of the other configuration options at this time, so we still support `--contracts.config-file-path` and `--contracts.config-json`. We can think about deprecating at least one of those. Don't think we need three ways to configure the same thing.
- For the CLI, I decided to not add a new option and instead to use a custom URL scheme `config://` (I also considered `env://`). This seemed less disruptive to the CLI codebase. I still think the custom scheme is kinda weird.

## Review notes

These configs get used a lot, with all the integration tests and images used in libxmtp. I've tested everything I can think of, but there may be uses of these flags I am not aware of.

Things I've done:

- Ran `./dev/up`​ 

- Ran `./dev/prune|run|run-2|run-gateway|test`

- CI ran the integration tests


neekolas:
* **#1248** <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/1248?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/xmtp/xmtpd/1248?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

## How to use the Graphite Merge Queue

Add either label to this PR to merge it via the merge queue:
- <kbd>Queue</kbd> - adds this PR to the back of the merge queue
- <kbd>Hotfix</kbd> - for urgent hot fixes, skip the queue and merge this PR next

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/xmtp?ref=merge-queue-instructions-in-stack-comment&prId=1248).

<h2></h2>
<sub>

An organization admin has enabled the [Graphite Merge Queue](https://graphite.dev/docs/graphite-merge-queue) in this repository.

</sub>

<sub>

Please do not merge from GitHub as this will restart CI on PRs being processed by the merge queue.

</sub>

<h2></h2>

dev/docker/docker-compose-register.yml
```diff
@@ -50,16 +48,15 @@ services:
     profiles: ["dual"]
     env_file:
       - ../local.env
-    volumes:
-      - ../environments/anvil.json:/cfg/anvil.json
     command:
       [
-        "--config-file=/cfg/anvil.json",
+        "--config-file=config://anvil",
```
neekolas: The `config://` in the `config-file` URL is a little weird to me. Open to adding a second flag to the CLI to specify an env. Or maybe a better name? I toyed with `env://`

pkg/config/environments/testnet-staging.json
```diff
@@ -0,0 +1,23 @@
+{
```
neekolas: This came straight out of terraform, along with `testnet.json`

pkg/config/environments/environments.go
```diff
@@ -0,0 +1,49 @@
...
+
+type SmartContractEnvironment string
+
+const (
+	Anvil          SmartContractEnvironment = "anvil"
+	Testnet        SmartContractEnvironment = "testnet"
+	TestnetStaging SmartContractEnvironment = "testnet-staging"
+)
+
+func (s *SmartContractEnvironment) UnmarshalFlag(value string) error {
```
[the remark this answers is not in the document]
mkysel: correct. That is why we have the manual types. It seems that `ignore` does not do its job

PR #1272, opened 2025-10-23
Speed up Vector Clock selection

mkysel:
`SelectVectorClock` never finishes. Even with a proper secondary index, the query is just too complicated. I've played around with a few alternatives and having a dedicated table that contains exactly 1 record seems like the best way forward.

The trigger keeps the table updated and transactional.

The time went down to sub 100ms.

If the best attempt fails, the consequence is that the `publish worker` will have to do some useless work when spinning up.

### DIFF
--- a/pkg/migrations/00020_add_latest_envelopes.up.sql
+++ b/pkg/migrations/00020_add_latest_envelopes.up.sql
@@ -6,12 +6,13 @@ CREATE TABLE gateway_envelopes_latest (
 
 CREATE OR REPLACE FUNCTION update_latest_envelope() RETURNS trigger AS $$
 BEGIN
-    INSERT INTO gateway_envelopes_latest
+    INSERT INTO gateway_envelopes_latest as g
     VALUES (NEW.originator_node_id, NEW.originator_sequence_id, NEW.gateway_time)
     ON CONFLICT (originator_node_id)
         DO UPDATE
         SET originator_sequence_id = EXCLUDED.originator_sequence_id,
-            gateway_time = EXCLUDED.gateway_time;
+            gateway_time = EXCLUDED.gateway_time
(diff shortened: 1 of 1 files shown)


pkg/migrations/00020_add_latest_envelopes.up.sql
```diff
@@ -0,0 +1,37 @@
...
+        DO UPDATE
+        SET originator_sequence_id = EXCLUDED.originator_sequence_id,
+            gateway_time = EXCLUDED.gateway_time;
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER gateway_latest_upd
+    AFTER INSERT ON gateway_envelopes
+    FOR EACH ROW EXECUTE FUNCTION update_latest_envelope();
```
tylerhawkes: Doing this for every row seems kind of heavy. Is there a way to do this once for each transaction from the app?

pkg/migrations/00020_add_latest_envelopes.up.sql
```diff
@@ -0,0 +1,37 @@
...
+);
+
+CREATE OR REPLACE FUNCTION update_latest_envelope() RETURNS trigger AS $$
+BEGIN
+    INSERT INTO gateway_envelopes_latest
+    VALUES (NEW.originator_node_id, NEW.originator_sequence_id, NEW.gateway_time)
+    ON CONFLICT (originator_node_id)
+        DO UPDATE
+        SET originator_sequence_id = EXCLUDED.originator_sequence_id,
+            gateway_time = EXCLUDED.gateway_time;
```
tylerhawkes: These can move backward. Probably want `greatest(OLD.gateway_time, EXCLUDED.gateway_time)` to enforce that they only move forward.

neekolas (approved):

PR #1338, opened 2025-11-19
PayerReports-3: Do not increment usage for reserved topic envelopes

fbac:
Root cause for "message is not the last of the minute" errors

### DIFF
--- a/pkg/db/gateway_envelope.go
+++ b/pkg/db/gateway_envelope.go
@@ -78,45 +78,6 @@ func InsertGatewayEnvelopeAndIncrementUnsettledUsage(
 	)
 }
 
-// InsertGatewayEnvelope inserts a gateway envelope into the database.
-// It does not increment the unsettled usage and congestion counters.
-// Used exclusively for reserved topics.
-//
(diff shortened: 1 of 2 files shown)


pkg/db/gateway_envelope.go
```diff
@@ -78,6 +78,45 @@ func InsertGatewayEnvelopeAndIncrementUnsettledUsage(
...
+// InsertGatewayEnvelope inserts a gateway envelope into the database.
+// It does not increment the unsettled usage and congestion counters.
+// Used exclusively for reserved topics.
+//
+// This function runs inside a managed transaction created by RunInTxWithResult().
+//
+// Steps:
+//  1. Calls InsertGatewayEnvelopeWithChecksTransactional() to insert the envelope,
+//     automatically creating any missing partitions if needed.
+func InsertGatewayEnvelope(
```
mkysel: you do not need this function. There is no need for a transaction. You are only executing a single statement. You should use `InsertGatewayEnvelopeWithChecksStandalone`

PR #1345, opened 2025-11-20
Fix sorted stream system invariant

mkysel:
### Enforce per-originator monotonic sequence ordering and abort on out-of-order envelopes in `message.Service.QueryEnvelopes` and `message.startSubscribeWorker`, and remove `gateway_time` from DB ORDER BY to fix sorted stream system invariant
Add per-originator last-seen checks that `logger.Fatal` on decreasing `OriginatorSequenceID`, drop `gateway_time` from envelope query ordering, and validate streams only by sequence progression without timestamps.

#### 📍Where to Start
Start with the per-originator ordering guard in `message.Service.QueryEnvelopes` in [service.go](https://github.com/xmtp/xmtpd/pull/1345/files#diff-1f687c3f8e69fec5867ded411d50bda02bef3661181fa7422b9dc66b7b57f3c2), then review DB ordering changes in [envelopes_v2.sql](https://github.com/xmtp/xmtpd/pull/1345/files#diff-26ca3ff0889ad4b2aa77e6edc7cafa970393fe952969c76c781fb136df2c07e3).

#### Changes since #1345 opened

- Replaced cursor-based sequence tracking with direct uint64 sequence ID tracking [5f0c1ef]
- Updated test infrastructure to accommodate new sequence tracking approach [5f0c1ef]

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 5f0c1ef. 4 files reviewed, 7 issues evaluated, 7 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>pkg/api/message/service.go — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 361](https://github.com/xmtp/xmtpd/blob/5f0c1efc18c6227569f032ab4de8ceee94bfc1e4/pkg/api/message/service.go#L361): Request handler `QueryEnvelopes` hard-terminates the process via `logger.Fatal(...)` when it detects an out-of-order sequence for a given originator (lines 361–369). Using `zap.Logger.Fatal` typically calls `os.Exit(1)`, abruptly killing the server mid-request. This violates the API contract (the unary handler should return a defined error/result), prevents cleanup of in-flight resources, and can drop other concurrent requests/subscriptions. Replace the hard crash with a controlled error path (e.g., return `connect.NewError` or map to an internal error) and add instrumentation/alerting. If a strict invariant must be enforced, ensure it is validated at ingestion time or guarded pre-effect. <b>[ Low confidence ]</b>
</details>

<details>
<summary>pkg/api/message/subscribe_worker.go — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 181](https://github.com/xmtp/xmtpd/blob/5f0c1efc18c6227569f032ab4de8ceee94bfc1e4/pkg/api/message/subscribe_worker.go#L181): Background subscribe worker `pollableQuery` hard-terminates the process via `logger.Fatal(...)` when detecting out-of-order sequences (lines 181–189). This executes in the subscription polling goroutine and will exit the whole process, leaving channels, subscriptions, and DB resources without orderly teardown. Prefer logging an error and surfacing the fault to health checks or metrics, then quarantining the offending data or initiating controlled recovery. If a strict invariant must be enforced, validate and enforce ordering at insertion time and fail the poll with an error instead of terminating. <b>[ Low confidence ]</b>
</details>

<details>
<summary>pkg/sync/originator_stream.go — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 97](https://github.com/xmtp/xmtpd/blob/5f0c1efc18c6227569f032ab4de8ceee94bfc1e4/pkg/sync/originator_stream.go#L97): Possible panic on send to a closed channel: `listen()` sends parsed envelopes to `s.writeQueue` (`s.writeQueue <- parsedEnv`, line 97). In the subscription goroutine, `writeQueue` is closed via `defer close(writeQueue)` when the goroutine exits on `ctx.Done`. There’s no coordination ensuring the producer (`listen`) has stopped sending before the channel is closed, so on shutdown a race can occur where `writeQueue` is closed while `listen` is still processing and sending, causing a panic. Guard with context cancellation, a draining protocol, or ensure channel closure happens after the listener goroutine has fully stopped. <b>[ Low confidence ]</b>
- [line 148](https://github.com/xmtp/xmtpd/blob/5f0c1efc18c6227569f032ab4de8ceee94bfc1e4/pkg/sync/originator_stream.go#L148): In `validateEnvelope`, envelopes with an `OriginatorNodeID` that does not equal `s.node.NodeID` are treated as errors and dropped. During migration, `setupStream` subscribes to multiple `OriginatorNodeIds` (including `GroupMessageOriginatorID`, `WelcomeMessageOriginatorID`, `KeyPackagesOriginatorID`), so this check will discard all migrated envelopes from those IDs, defeating the migration path and causing permanent data loss. Consider either instantiating distinct streams per originator or relaxing this check when multiple `OriginatorNodeIds` are requested. <b>[ Low confidence ]</b>
- [line 164](https://github.com/xmtp/xmtpd/blob/5f0c1efc18c6227569f032ab4de8ceee94bfc1e4/pkg/sync/originator_stream.go#L164): Potential integer overflow when logging sequence ID: `utils.SequenceIDField(int64(env.OriginatorSequenceID()))` casts `uint64` to `int64` and can overflow for values > `math.MaxInt64`, producing incorrect log values. Use `zap.Uint64` or change `utils.SequenceIDField` to accept `uint64`. <b>[ Previously rejected ]</b>
</details>

<details>
<summary>pkg/sync/sync_worker.go — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 385](https://github.com/xmtp/xmtpd/blob/5f0c1efc18c6227569f032ab4de8ceee94bfc1e4/pkg/sync/sync_worker.go#L385): `lastSequenceId` is derived as a single scalar across potentially multiple `OriginatorNodeIds` in `setupStream`, and later used in `validateEnvelope` for order checking. When migration enables multiple originators, the loop (lines 384–389) may leave `lastSequenceId` set to the last matching row, which can correspond to a different originator than the envelope being validated. This yields incorrect `expected_sequence_id` checks and updates, causing spurious out-of-order errors and incorrect sequence tracking. <b>[ Out of scope ]</b>
- [line 392](https://github.com/xmtp/xmtpd/blob/5f0c1efc18c6227569f032ab4de8ceee94bfc1e4/pkg/sync/sync_worker.go#L392): `newOriginatorStream` is called with `s.ctx` instead of the per-registration `ctx` passed to `setupStream`. As a result, `originatorStream.listen()` observes the worker-level context, not the registration-specific context. Cancellation of the registration (e.g., node config changes) won’t trigger the `ctx.Done()` path in `listen`; termination then relies on gRPC stream cancellation errors propagating from the separate context used in `SubscribeEnvelopes`. This mismatched context can delay shutdown and complicate control flow and logging. <b>[ Low confidence ]</b>
</details>

</details>

### DIFF
--- a/pkg/sync/originator_stream.go
+++ b/pkg/sync/originator_stream.go
@@ -16,24 +16,20 @@ import (
 	"go.uber.org/zap"
 )
 
-type cursor struct {
-	sequenceID uint64
-}
-
 type originatorStream struct {
-	ctx        context.Context
-	logger     *zap.Logger
-	node       *registry.Node
-	cursor     *cursor
-	stream     message_api.ReplicationApi_SubscribeEnvelopesClient
-	writeQueue chan *envUtils.OriginatorEnvelope
+	ctx            context.Context
+	logger         *zap.Logger
+	node           *registry.Node
+	lastSequenceId uint64
+	stream         message_api.ReplicationApi_SubscribeEnvelopesClient
+	writeQueue     chan *envUtils.OriginatorEnvelope
 }
 
 func newOriginatorStream(
 	ctx context.Context,
 	logger *zap.Logger,
 	node *registry.Node,
-	cursor *cursor,
+	lastSequenceId uint64,
 	stream message_api.ReplicationApi_SubscribeEnvelopesClient,
 	writeQueue chan *envUtils.OriginatorEnvelope,
 ) *originatorStream {
@@ -43,10 +39,10 @@ func newOriginatorStream(
 			utils.OriginatorIDField(node.NodeID),
 			utils.NodeHTTPAddressField(node.HTTPAddress),
 		),
-		node:       node,
-		cursor:     cursor,
-		stream:     stream,
-		writeQueue: writeQueue,
+		node:           node,
+		lastSequenceId: lastSequenceId,
+		stream:         stream,
+		writeQueue:     writeQueue,
 	}
 }
 
@@ -161,24 +157,17 @@ func (s *originatorStream) validateEnvelope(
 	metrics.EmitSyncLastSeenOriginatorSequenceID(env.OriginatorNodeID(), env.OriginatorSequenceID())
 	metrics.EmitSyncOriginatorReceivedMessagesCount(env.OriginatorNodeID(), 1)
 
-	var lastSequenceID uint64 = 0
-	if s.cursor != nil {
-		lastSequenceID = s.cursor.sequenceID
-	}
-
-	if env.OriginatorSequenceID() != lastSequenceID+1 {
+	if env.OriginatorSequenceID() != s.lastSequenceId+1 {
 		s.logger.Error(
 			"received out-of-order envelope",
 			utils.OriginatorIDField(env.OriginatorNodeID()),
 			utils.SequenceIDField(int64(env.OriginatorSequenceID())),
-			zap.Uint64("expected_sequence_id", lastSequenceID+1),
+			zap.Uint64("expected_sequence_id", s.lastSequenceId+1),
 		)
 	}
 
-	if env.OriginatorSequenceID() > lastSequenceID {
-		s.cursor = &cursor{
-			sequenceID: env.OriginatorSequenceID(),
-		}
+	if env.OriginatorSequenceID() > s.lastSequenceId {
+		s.lastSequenceId = env.OriginatorSequenceID()
 	}
 
 	// Validate that there is a valid payer signature
--- a/pkg/sync/originator_stream_test.go
+++ b/pkg/sync/originator_stream_test.go
@@ -84,7 +84,7 @@ func newTestOriginatorStream(
 	t *testing.T,
 	node *registry.Node,
 	stream message_api.ReplicationApi_SubscribeEnvelopesClient,
-	cursor *cursor,
+	lastSequenceId uint64,
 	writeQueue chan *envUtils.OriginatorEnvelope,
 ) *originatorStream {
 	log := testutils.NewLog(t)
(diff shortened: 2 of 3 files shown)


pkg/sync/originator_stream.go
```diff
@@ -17,8 +17,7 @@ import (
 )
 
 type cursor struct {
-	sequenceID  uint64
-	timestampNS int64
+	sequenceID uint64
 }
```
fbac: We don't really need this struct anymore. An *uint64 in the originatorStream cursor would be more straight forward.

fbac (approved):

pkg/api/message/subscribe_worker.go
```diff
@@ -178,9 +178,17 @@ func startSubscribeWorker(
 			nodeID := uint32(env.OriginatorNodeID)
 			seqID := uint64(env.OriginatorSequenceID)
 
-			if current, ok := lastSeen[nodeID]; !ok || seqID > current {
-				lastSeen[nodeID] = seqID
+			if last, ok := lastSeen[nodeID]; ok && seqID < last {
```
neekolas: I do worry about abuse potential here.  
  
Node A can publish bad payloads and completely take down Node B. It's fine for testing but I worry about this in prod.  
  
A middle-ground would be completely shutting down the stream to the node that breaks this invariant, and we fire a metric that alerts on the first call.

neekolas: Oh, I’m sorry. I thought this was on the read from the stream. Agree these never should get into the DB in the first place.

PR #1437, opened 2025-12-11
Migrator: track commits and group messages separately

fbac:
### Track commit messages separately from group messages and route commit messages to blockchain while writing non-commit group messages to the database in `pkg/migrator`
Introduce a dedicated `CommitMessage` type, reader, and originator ID, split `group_messages` by `is_commit`, route commit messages to blockchain and non-commit group messages to DB, add migration tracker entries for `commit_messages`, and make `Migrator.running` atomic. Update transformers to set retention by destination and adjust tests accordingly.

#### 📍Where to Start
Start with routing and lifecycle changes in `Migrator` within [migrator.go](https://github.com/xmtp/xmtpd/pull/1437/files#diff-ccb0d31b1e7491838a2e4c7f2f6523eed971d6f18d4c04f006e14e2767671693), then review transformation logic in [transformer.go](https://github.com/xmtp/xmtpd/pull/1437/files#diff-c76d70079716ba42e22c5b5031c1960504cccbca874aadb118385ca67341cf62) and the new commit message reader in [reader.go](https://github.com/xmtp/xmtpd/pull/1437/files#diff-5afb69a12b4d409ea45341dca6b796f48dc578ab56ad078142152b5628bf0828).

#### Changes since #1437 opened

- Added start date filtering to migration readers [9fc0436]
- Moved source sequence ID metrics emission from worker to readers with periodic polling [9fc0436]
- Consolidated destination sequence ID metrics from separate blockchain and database metrics to single metric with table label [9fc0436]

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 9fc0436. 8 files reviewed, 12 issues evaluated, 8 issues filtered, 1 comment posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>pkg/config/migrator.go — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 17](https://github.com/xmtp/xmtpd/blob/9fc0436ce0b0b7a448de01c8bc78c118720fc669/pkg/config/migrator.go#L17): The `time.Time` type added for `StartDate` field may not be supported by the flag parsing library (likely `go-flags`) without a custom unmarshaler. Standard `go-flags` does not have built-in support for parsing `time.Time` values, which would cause a runtime panic or error when attempting to parse the `--start-date` flag or `XMTPD_MIGRATION_START_DATE` environment variable. Either a custom type implementing the `Unmarshaler` interface is needed, or the library must be extended to handle `time.Time`. <b>[ Low confidence ]</b>
</details>

<details>
<summary>pkg/migrator/migrator.go — 0 comments posted, 5 evaluated, 5 filtered</summary>

- [line 182](https://github.com/xmtp/xmtpd/blob/9fc0436ce0b0b7a448de01c8bc78c118720fc669/pkg/migrator/migrator.go#L182): If `cfg.options.StartDate` is the zero value of `time.Time` (parsing failure or unset), calling `.Unix()` or `.UnixNano()` will return negative values representing year 1 CE. This could cause unexpected filtering behavior - potentially returning all records or no records depending on database column values and comparisons. <b>[ Low confidence ]</b>
- [line 184](https://github.com/xmtp/xmtpd/blob/9fc0436ce0b0b7a448de01c8bc78c118720fc669/pkg/migrator/migrator.go#L184): The SQL queries for `NewGroupMessageReader`, `NewCommitMessageReader`, `NewInboxLogReader`, and `NewWelcomeMessageReader` all reference `$3` for the `startDate` filter, but `NewKeyPackageReader` passes `0` as `startDate` to `NewDBReader` while its query only uses `$1` and `$2`. If `DBReader.Fetch` always binds three parameters (`lastID`, `limit`, `startDate`), the `KeyPackageReader` query will fail at runtime with a PostgreSQL error because the query only expects 2 parameters. <b>[ Low confidence ]</b>
- [line 242](https://github.com/xmtp/xmtpd/blob/9fc0436ce0b0b7a448de01c8bc78c118720fc669/pkg/migrator/migrator.go#L242): The `Stop()` method now returns an error when called on a non-running migrator, but the example usage in `server.go` logs the error with `s.logger.Error()`. The previous behavior returned `nil` when not running, making `Stop()` idempotent. The new behavior breaks this idempotency - if `Stop()` is called twice (e.g., during shutdown cleanup), the second call will return an error that gets logged as a failure, even though this may be expected behavior. This is a semantic change that could cause spurious error logs or break callers expecting idempotent shutdown. <b>[ Low confidence ]</b>
- [line 284](https://github.com/xmtp/xmtpd/blob/9fc0436ce0b0b7a448de01c8bc78c118720fc669/pkg/migrator/migrator.go#L284): The `cleanupInflight` function at line 279-290 has a `default` case that silently ignores the semaphore release when it's full. Combined with the ID mismatch bug, if the semaphore token was acquired for a record but cleanup is called with a different ID (the OriginatorSequenceID), the delete succeeds on a non-existent key, but the semaphore may already be full from a previous buggy release, causing the actual token to be lost and eventually starving the semaphore. <b>[ Out of scope ]</b>
- [line 516](https://github.com/xmtp/xmtpd/blob/9fc0436ce0b0b7a448de01c8bc78c118720fc669/pkg/migrator/migrator.go#L516): The `inflight` map is keyed by `record.GetID()` when records are added (line 418), but lookups and deletions in the writer use `env.OriginatorSequenceID()` (lines 516, 550). If these IDs differ (which is likely since `GetID()` is the source database primary key and `OriginatorSequenceID()` is assigned during transformation), the `inflight` map entries will never be found or deleted, causing: (1) E2E latency metrics to never be emitted since `exists` will always be false, (2) memory leak as entries accumulate in the `inflight` map, and (3) potential semaphore starvation since `cleanupInflight` deletes non-existent keys. <b>[ Filtered as negative feedback ]</b>
</details>

<details>
<summary>pkg/migrator/reader.go — 1 comment posted, 5 evaluated, 1 filtered</summary>

- [line 44](https://github.com/xmtp/xmtpd/blob/9fc0436ce0b0b7a448de01c8bc78c118720fc669/pkg/migrator/reader.go#L44): In `Fetch()`, the height query is executed when `time.Now().Minute()%10 == 0`, which means it runs for the entire minute (not just once every 10 minutes). If `Fetch()` is called multiple times during that minute, each call will execute the height query, potentially causing unnecessary database load. Additionally, there's no synchronization to prevent concurrent metric updates. <b>[ Low confidence ]</b>
</details>

<details>
<summary>pkg/migrator/transformer.go — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 373](https://github.com/xmtp/xmtpd/blob/9fc0436ce0b0b7a448de01c8bc78c118720fc669/pkg/migrator/transformer.go#L373): In `transformGroupMessage`, line 373 accesses `groupMessage.ShouldPush.Bool` directly without checking `groupMessage.ShouldPush.Valid`. Since `ShouldPush` is a `sql.NullBool`, when `Valid` is `false` (indicating a NULL database value), the code silently uses the zero value (`false`) for `ShouldPush`. This may not be the intended behavior - a NULL value might semantically mean "use default" or "true" rather than explicitly `false`. The same issue applies to potential usage patterns with other `sql.NullBool` fields if they follow this pattern. <b>[ Low confidence ]</b>
</details>

</details>


fbac:
Closes #1431

PR #1455, opened 2025-12-16
Migrator: identity update batches

fbac:
### Batch identity updates in the migrator and start dedicated workers per table with timed (250ms) or 100KB flush in `migrator.Migrator.startInboxLogWorker` in [migrator.go](https://github.com/xmtp/xmtpd/pull/1455/files#diff-ccb0d31b1e7491838a2e4c7f2f6523eed971d6f18d4c04f006e14e2767671693)
Introduce `migrator.Worker` and replace the monolithic pipeline with dedicated workers per table, add batched identity update publishing with per-item fallback and dead-lettering, and add `FailureBlockchainUndetermined` along with an `IdentityUpdateBatch` utility. See [migrator.go](https://github.com/xmtp/xmtpd/pull/1455/files#diff-ccb0d31b1e7491838a2e4c7f2f6523eed971d6f18d4c04f006e14e2767671693), [worker.go](https://github.com/xmtp/xmtpd/pull/1455/files#diff-98d986f4e853aaa039c51a1e336347facf3ecc8a9c45ff146d1dcfd8883c94b8), and [writer.go](https://github.com/xmtp/xmtpd/pull/1455/files#diff-79de0174d18bbb97eb4c62be1eeec1e2132ff3060f5beafb3b8721044791b258).

#### 📍Where to Start
Start with `migrator.Migrator.Start` and the worker startup helpers in [migrator.go](https://github.com/xmtp/xmtpd/pull/1455/files#diff-ccb0d31b1e7491838a2e4c7f2f6523eed971d6f18d4c04f006e14e2767671693), then review batching logic in `migrator.Worker.startBlockchainWriterIdentityUpdateBatches` and `flushIdentityUpdatesBatch` in [worker.go](https://github.com/xmtp/xmtpd/pull/1455/files#diff-98d986f4e853aaa039c51a1e336347facf3ecc8a9c45ff146d1dcfd8883c94b8) and [writer.go](https://github.com/xmtp/xmtpd/pull/1455/files#diff-79de0174d18bbb97eb4c62be1eeec1e2132ff3060f5beafb3b8721044791b258).

#### Changes since #1455 opened

- Updated logging in identity update batch processing within the `migrator.Worker` type [f2845c1]

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized f2845c1. 6 files reviewed, 19 issues evaluated, 16 issues filtered, 1 comment posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>pkg/migrator/migrator.go — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 258](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/migrator.go#L258): The `Worker` instances created in `startKeyPackagesWorker`, `startWelcomeMessagesWorker`, `startGroupMessagesWorker`, `startCommitMessagesWorker`, and `startInboxLogWorker` use their own `w.wg` `sync.WaitGroup` for goroutine tracking (via `tracing.GoPanicWrap(ctx, &w.wg, ...)`), but `Stop()` only waits on `m.wg`. Since the workers are local variables that are not stored on the `Migrator` struct, and their `WaitGroup` instances are never linked to `m.wg`, calling `m.wg.Wait()` in `Stop()` returns immediately without waiting for worker goroutines to finish. This causes a race condition where the `reader` and `writer` database connections are closed while worker goroutines may still be actively using them. <b>[ Out of scope ]</b>
- [line 274](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/migrator.go#L274): The `Worker` instances created in each `start*Worker()` method are local variables that are never stored on the `Migrator` struct. After these methods return, there is no reference to the workers, making it impossible to properly manage their lifecycle (e.g., call any cleanup methods on them or access their `WaitGroup` for shutdown coordination). <b>[ Previously rejected ]</b>
</details>

<details>
<summary>pkg/migrator/types.go — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 222](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/types.go#L222): The new `FailureBlockchainUndetermined` constant is added but `ShouldRetry()` method's switch statement does not have an explicit case for it. It falls through to the `default` case which returns `false`. If this failure reason should actually trigger retries (which seems plausible given its 'undetermined' nature), this would be a bug. The caller in `flushIdentityUpdatesBatch` inserts to dead letter box when individual retries fail, so this affects whether blockchain undetermined errors are retried. <b>[ Low confidence ]</b>
</details>

<details>
<summary>pkg/migrator/worker.go — 1 comment posted, 9 evaluated, 7 filtered</summary>

- [line 119](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/worker.go#L119): Using `logger.Fatal` when `GetMigrationProgress` fails will terminate the entire process via `os.Exit(1)`. This seems overly aggressive for what could be a transient database error. Other errors in the same function use `logger.Error` and retry logic instead. <b>[ Already posted ]</b>
- [line 135](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/worker.go#L135): Using direct equality comparison `err == sql.ErrNoRows` instead of `errors.Is(err, sql.ErrNoRows)`. If `reader.Fetch` wraps the `sql.ErrNoRows` error (common in database abstraction layers), this comparison will fail, causing the error to fall through to the default case and be treated as a generic error with `sleepTimeOnError` instead of `sleepTimeOnNoRows`. <b>[ Already posted ]</b>
- [line 195](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/worker.go#L195): Semaphore leak when context is cancelled after acquiring semaphore but before sending record. At line 191, the semaphore is acquired via `<-w.sem`. If context cancellation is detected at line 195-196, the function returns without releasing the semaphore (no call to `cleanupInflight`). This will permanently consume one semaphore slot, reducing throughput over time. <b>[ Already posted ]</b>
- [line 252](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/worker.go#L252): The oversized message check for `InboxLogOriginatorID` was removed in the diff but kept for `CommitMessageOriginatorID`. Looking at the diff, lines 252-280 that contained the `if totalSize > maxChainMessageSize` check for identity updates were deleted. This means oversized identity updates will now be sent to the blockchain without the pre-check, potentially causing `InvalidPayloadSize()` errors that are only caught reactively rather than proactively handled with dead letter box insertion upfront. <b>[ Low confidence ]</b>
- [line 267](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/worker.go#L267): Inflight map entry and semaphore leak when context is cancelled after successfully transforming a record. If transform succeeds at line 240 and context is cancelled at line 267, the function returns without calling `cleanupInflight`. The record remains in `w.inflight` and the semaphore slot is never released. <b>[ Already posted ]</b>
- [line 541](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/worker.go#L541): Inside `triggerBatchFlush`, when the batch flush fails (line 521), the function logs the error, emits metrics, cleans up inflight entries, resets the batch, and then calls `return` at line 541. This `return` exits `triggerBatchFlush` only, not the outer goroutine loop. When called from the ticker case (line 592) or channel close case (line 603), the goroutine continues or exits respectively, which is likely correct. However, when called from the batch size threshold case (line 653), after a flush error the loop continues and may immediately receive another envelope, potentially causing rapid repeated failures without any backoff mechanism. <b>[ Low confidence ]</b>
- [line 685](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/worker.go#L685): In `cleanupInflight`, the semaphore release uses a `default` case to silently ignore double-release attempts. While the comment indicates this is intentional, if there's a bug elsewhere causing double-releases, it will be silently swallowed without any logging, making debugging difficult. Consider adding a log or metric when the default case is hit. <b>[ Low confidence ]</b>
</details>

<details>
<summary>pkg/migrator/writer.go — 0 comments posted, 7 evaluated, 6 filtered</summary>

- [line 27](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/writer.go#L27): The validation that checked whether `env.OriginatorNodeID()` maps to a valid table name was removed. Previously the code validated the envelope's originator ID against `originatorIDToTableName` and returned an error for invalid IDs. Now `w.tableName` is used directly without verifying the envelope actually belongs to this worker's table, which could cause migration progress to be tracked incorrectly if an envelope with an unexpected originator ID is processed. <b>[ Low confidence ]</b>
- [line 138](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/writer.go#L138): The `totalSize` variable calculated at line 138 is used for the `CommitMessageOriginatorID` case but is never used for the `InboxLogOriginatorID` case since the size check was removed. This is dead code that adds unnecessary computation for identity updates. <b>[ Low confidence ]</b>
- [line 246](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/writer.go#L246): The proactive size check (`if totalSize > maxChainMessageSize`) was removed from the `InboxLogOriginatorID` case (identity updates), while it remains for `CommitMessageOriginatorID` (commit messages). This means oversized identity updates will attempt blockchain publication and only be caught by the `InvalidPayloadSize()` error handler (lines 271-290), resulting in unnecessary blockchain calls that will fail. The asymmetry in handling between the two message types could indicate an incomplete refactoring. <b>[ Already posted ]</b>
- [line 380](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/writer.go#L380): In `flushIdentityUpdatesBatch`, after the batch retry fails and individual retries are attempted, if a `context.Canceled` or `context.DeadlineExceeded` error occurs during an individual retry (inside the `for identityUpdate := range batch.All()` loop), the error is checked but the function continues to the next iteration via `continue` instead of returning early. The `errors.Is` check only happens on the initial batch retry error (line 357-359), not on individual retry errors. This means if context is cancelled during individual processing, remaining items will still be attempted. <b>[ Already posted ]</b>
- [line 461](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/writer.go#L461): In `bootstrapIdentityUpdates`, if an empty `IdentityUpdateBatch` is passed, the function will call `BootstrapIdentityUpdates` with empty slices, then call `UpdateMigrationProgress` with `LastMigratedID: 0` (since `LastSequenceID()` returns 0 for empty batches). This could incorrectly reset migration progress to 0, causing previously migrated records to be re-processed on restart. <b>[ Already posted ]</b>
- [line 550](https://github.com/xmtp/xmtpd/blob/f2845c125e9fe49a8ba41bc09f6255400ba93767/pkg/migrator/writer.go#L550): The `retry` function has no maximum retry limit. When `err.ShouldRetry()` returns true, it will continue retrying indefinitely until either the context is cancelled or a non-recoverable error occurs. For persistent recoverable errors (e.g., database connectivity issues that last a long time), this could cause the worker to spin indefinitely on a single message, blocking progress on the channel. <b>[ Low confidence ]</b>
</details>

</details>

### DIFF
--- a/pkg/migrator/worker.go
+++ b/pkg/migrator/worker.go
@@ -502,7 +502,7 @@ func (w *Worker) startBlockchainWriterIdentityUpdateBatches(ctx context.Context)
 				lastSequenceID := identityUpdateBatch.LastSequenceID()
 
 				logger.Info(
-					"flushing identity update batch",
+					"publishing identity update batch",
 					zap.Int("length", identityUpdateBatch.Len()),
 					zap.Uint64("last_sequence_id", lastSequenceID),
 				)
@@ -522,7 +522,7 @@ func (w *Worker) startBlockchainWriterIdentityUpdateBatches(ctx context.Context)
 					logger.Error(
 						"failed to flush identity update batch",
 						zap.Int("length", identityUpdateBatch.Len()),
-						zap.Uint64("last_sequence_id", lastSequenceID),
+						utils.SequenceIDField(int64(lastSequenceID)),
 						zap.Error(err),
 					)
 
--- a/pkg/migrator/writer.go
+++ b/pkg/migrator/writer.go
@@ -472,6 +472,7 @@ func (w *Worker) bootstrapIdentityUpdates(
 
 	w.logger.Debug(
 		"published identity update batch",
+		zap.Int("length", batch.Len()),
 		utils.SequenceIDField(int64(batch.LastSequenceID())),
 	)
 


pkg/migrator/migrator.go
```diff
@@ -594,7 +618,55 @@ func (m *Migrator) migrationWorker(tableName string) {
...
+								logger.Error(
+									"failed to prepare identity update envelope",
+									zap.Error(err),
+								)
+								return
+							}
+
+							inboxID, err := utils.ParseInboxID(identifier)
+							if err != nil {
+								logger.Error("failed to parse inbox ID", zap.Error(err))
```
Maelkum: Perhaps also log the identifier so we can see which one we failed to parse?

pkg/migrator/migrator.go
```diff
@@ -594,7 +618,55 @@ func (m *Migrator) migrationWorker(tableName string) {
...
+
+							// Add to batch.
+							identityUpdateBatch.Add(
+								inboxID,
+								clientEnvelopeBytes,
+								sequenceID,
+							)
+
+							// Check if batch should be flushed, based on size threshold.
+							if identityUpdateBatch.Size() >= maxBatchSize {
```
Maelkum: Potentially invert and return `if identityUpdateBatch.Size() < maxBatchSize` ?

pkg/migrator/writer.go
```diff
@@ -353,6 +321,183 @@ func (m *Migrator) insertOriginatorEnvelopeBlockchain(
 	return nil
 }
 
+func (m *Migrator) prepareClientEnvelope(
+	env *envelopes.OriginatorEnvelope,
+	tableName string,
+) (clientEnvelopeBytes []byte, identifier []byte, sequenceID uint64, totalSize int, err error) {
+	sequenceID = env.OriginatorSequenceID()
+	identifier = env.TargetTopic().Identifier()
```
Maelkum: Curious - the sequenceID and identifier come from the function inputs (envelope). Does the function have to return them? They're known before the function call already, we could have two return values less.

pkg/migrator/worker.go
```diff
@@ -0,0 +1,688 @@
...
+							return w.wrtrQueries.GetMigrationProgress(ctx, w.tableName)
+						},
+					)
+					if err != nil {
+						metrics.EmitMigratorReaderError(
+							"migration_tracker",
+							err.Error(),
+						)
+
+						logger.Fatal("failed to get migration progress", zap.Error(err))
```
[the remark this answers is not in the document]
fbac: If the migrator can't get the migration progress, the process should halt immediately and be fixed.

PR #1465, opened 2025-12-17
V3 Compatibility - Return nothing on empty filters

### LINKED ISSUE #1440: Bug: dont return everything for streams who subscribe with empty topics
### Describe the bug

libxmtp expects a stream that is subscribed with an empty topic list to return nothing. https://github.com/xmtp/libxmtp/issues/2924 will fix this on the client but would be nice to have a backend fix as well

### Expected behavior

_No response_

### Steps to reproduce the bug

_No response_

mkysel:
Fixes #1440

### DIFF
--- a/pkg/api/message/subscribe_test.go
+++ b/pkg/api/message/subscribe_test.go
@@ -163,7 +163,13 @@ func validateUpdates(
 
 			// Must not receive duplicates for the expected set.
 			_, dup := seen[k]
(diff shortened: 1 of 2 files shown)


insipx (approved):

neekolas:
@fbac @mkysel don't worry about that. I'll push up something small to satisfy push notification subscribers. This is the right choice for this endpoint.

Issue #1316, opened 2025-11-12
SQL Performance: Unfiltered scans all partitions


mkysel:
```
<head></head>
QUERY PLAN
--
Nested Loop  (cost=7.41..109343.99 rows=195106 width=729) (actual time=7571.213..13078.096 rows=1000 loops=1)
Buffers: shared hit=1458070
'  ->  Limit  (cost=6.99..585.99 rows=1000 width=51) (actual time=7571.145..13073.008 rows=1000 loops=1)
Buffers: shared hit=1454955
'        ->  Nested Loop Left Join  (cost=6.99..1509754.50 rows=2607478 width=51) (actual time=7571.143..13072.871 rows=1000 loops=1)
Join Filter: (m.originator_node_id = x.node_id)
Rows Removed by Join Filter: 39041648
Filter: (m.originator_sequence_id > COALESCE(y.seq_id, '0'::bigint))
Rows Removed by Filter: 7807330
Buffers: shared hit=1454955
'              ->  Merge Append  (cost=6.98..688398.79 rows=7822433 width=51) (actual time=0.492..4046.881 rows=7808330 loops=1)
Sort Key: m.gateway_time, m.originator_node_id, m.originator_sequence_id
Buffers: shared hit=1454955
'                    ->  Index Scan using gateway_envelopes_meta_o0_s0__gateway_time_originator_node__idx on gateway_envelopes_meta_o0_s0_1000000 m_1  (cost=0.42..16355.21 rows=278444 width=38) (actual time=0.030..111.049 rows=278390 loops=1)
Buffers: shared hit=68753
'                    ->  Index Scan using gateway_envelopes_meta_o0_s10_gateway_time_originator_node__idx on gateway_envelopes_meta_o0_s1000000_2000000 m_2  (cost=0.42..16918.21 rows=288260 width=38) (actual time=0.023..117.655 rows=288242 loops=1)
Buffers: shared hit=68740
'                    ->  Index Scan using gateway_envelopes_meta_o0_s20_gateway_time_originator_node__idx on gateway_envelopes_meta_o0_s2000000_3000000 m_3  (cost=0.42..16720.61 rows=284778 width=38) (actual time=0.029..114.026 rows=284655 loops=1)
Buffers: shared hit=67637
[...]

PR #1489, opened 2025-12-30
Subscribe worker queries envelopes per originator

Maelkum:
## Description

This PR changes how the subscribe worker queries envelopes - instead of using a single query which scans all partitions, worker now creates one subscription per known canonical originator. Worker also relies on the node registry and subscribes to new workers. Additionally, the registry has been extended to also spot removed nodes; worker also monitors this and removes subscription for such nodes.

Input channels (per-originator subscriptions) are funneled into a single channel which is processed. Limits on these were not touched though - it should probably be changed to be closer to the previous behavior.

## Misc 

Some cosmetic and low priority changes were done, some of which can be reverted:

### Test Log handling

Three environment variables added to control log outputs during tests. De

- `XMTP_TEST_LOG_LEVEL` - set log level for logger used during tests; default now is _debug_ which can get quite noisy
- `XMTP_TEST_DISABLE_STACK_TRACE` - we have negative tests, i.e. here's an invalid output => make sure we fail. Zap logs all messages with level _warn_ and above with a stack trace - which, again, makes log output quite noisy
- `XMTP_TEST_FILE_LOGGER` - instead of logging to stdout/stderr, save log messages as JSON in a log file, which can be easier to check and query using a tool like `jq`. Output file is `test_log_<testname>_hhmmss.json`

### Test API handling extended

Test API is extended in order to support adding more nodes to the registry, but other options can be added.

Reason is, the registry by default only has one node added, and for a more meaningful test we create envelopes from 4 nodes which we add to the registry, and monitor that our subscribe worker indeed picks them up.

Implements #1316

### DIFF
--- a/pkg/registry/node_registry_contract.go
+++ b/pkg/registry/node_registry_contract.go
@@ -159,26 +159,18 @@ func (s *SmartContractRegistry) refreshData() error {
 		return fmt.Errorf("could not load nodes from contract: %w", err)
 	}
 
-	var (
-		newNodes  []Node
-		seenNodes = make(map[uint32]struct{}) // Accounting of visited nodes
-	)
+	var newNodes []Node
 	for _, node := range fromContract {
 		// nodes realistically start at 100, but the contract fills the array with empty nodes
 		if !node.IsValidConfig {
 			continue
 		}
 
-		seenNodes[node.NodeID] = struct{}{}
-
 		existingValue, ok := s.nodes[node.NodeID]
 		if !ok {
 			// New node found
 			newNodes = append(newNodes, node)
-			continue
-		}
-
-		if !node.Equals(existingValue) {
+		} else if !node.Equals(existingValue) {
 			s.processChangedNode(node)
 		}
 	}


pkg/api/message/listener.go
```diff
@@ -0,0 +1,127 @@
+package message
```
Maelkum: This is unchanged, just moved to a separate file to remove clutter.

pkg/api/message/subscribe_worker.go
```diff
@@ -22,207 +22,96 @@ const (
...
+	if err != nil {
+		return nil, fmt.Errorf("could not get list of originators: %w", err)
+	}
+
+	var nodeIDs []uint32
+	for _, node := range nodes {
+		// TODO: Check - this seems like a sane thing to do but requires updating tests.
+		// if !node.IsCanonical {
+		// 	continue
+		// }
```
Maelkum: This check made sense to me but some tests were failing originally, so I commented it out.

If it indeed makes sense I can go around and update the tests and reinstate it

mkysel: you should never receive non-canonical nodes. What is the failure?

pkg/testutils/log.go
```diff
@@ -1,32 +1,90 @@
...
+	cfg.Level = level
+	cfg.DisableStacktrace = disableStackTrace
+
+	log, err := cfg.Build()
+	require.NoError(t, err)
+
+	return log
+}
+
+func NewJSONLog(t testing.TB) *zap.Logger {
```
mkysel: how are you going to access these logs for failed GH CI tests?

pkg/registry/interface.go
```diff
@@ -18,5 +18,6 @@ type NodeRegistry interface {
 	GetNode(uint32) (*Node, error)
 	OnNewNodes() <-chan []Node
 	OnChangedNode(uint32) <-chan Node
+	OnRemovedNodes() <-chan []uint32
```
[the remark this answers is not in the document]
Maelkum: Huh, okay. That would've made my life simpler. Will remove that code

pkg/api/message/envelope_list.go
```diff
@@ -0,0 +1,153 @@
...
+		logger:     logger,
+		store:      store,
+		cursor:     cursor,
+		mergedSubs: newFunnel[[]queries.SelectGatewayEnvelopesByOriginatorsRow](),
+	}
+
+	return s
+}
+
+func (s *subscriptionHandler) newSubscription(ctx context.Context, id uint32) (retErr error) {
```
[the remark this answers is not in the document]
Maelkum: I also dislike named returns and almost never use them.

However, I find it convenient for cases like this - if this function fails, do some cleanup (in this case cancel the child context).

Otherwise, on each return path you need to do it manually and tomorrow, someone will add another `if err != nil` branch and perhaps forget to include the `cancel()` call. 

Doing something like this handles it in a single place.

I can change it though.

pkg/api/message/subscribe_worker.go
Maelkum: Nothing serious, some tests are relaxed and create nodes that are not canonical (and corresponding envelopes). Suddenly subscribe worker is more strict and rejects them, and tests fail.

Just need to go through the tests and update the nodes with the `IsCanonical` flag set to true.

mkysel (approved):

PR #1530, opened 2026-01-12
Change envelope partition creation to allow less restrictive locking

Maelkum:
## Problem Description

Running tests sometimes fails with queries failing with an `40P01: deadlock detected` error. Failures can be reproduced by running tests involving creating/serving gateway envelopes in a loop.

## Root Cause Analysis

Enabled more detailed postgres logs, especially related to locks and locking failures. Logs revealed two queries that are stuck in a race - `ensure_gateway_parts` function call and `SelectGatewayEnvelopesByOriginators` query.

### Process A

Process executing `ensure_gateway_parts` wants an AccessExclusiveLock in order to create partitions which do not exist. This function call invokes 4 queries operating on two main tables - `gateway_envelopes_meta` and `gateway_envelope_blobs`. From Postgres documentation:

> ACCESS EXCLUSIVE (AccessExclusiveLock)
> 
> Conflicts with locks of all modes (ACCESS SHARE, ROW SHARE, ROW EXCLUSIVE, SHARE UPDATE EXCLUSIVE, SHARE, SHARE ROW EXCLUSIVE, EXCLUSIVE, and ACCESS EXCLUSIVE). **This mode guarantees that the holder is the only transaction accessing the table in any way.**

The reason for the exclusive lock is the query `CREATE TABLE X PARTITION OF Y` - this query requires exclusive access on the parent table.

> Note that creating a partition using PARTITION OF requires taking an ACCESS EXCLUSIVE lock on the parent partitioned table. Likewise, dropping a partition with DROP TABLE requires taking an ACCESS EXCLUSIVE lock on the parent table. 

### Process B

Process executing `SelectGatewayEnvelopesByOriginators` does a SELECT on the same two tables - `gateway_envelopes_meta` and `gateway_envelope_blobs`. The query does a select on the `meta` table, followed by a `JOIN` on `blobs`.

This query uses an AccessShareLock.

From the Postgres documentation:

> ACCESS SHARE (AccessShareLock)
> 
> Conflicts with the ACCESS EXCLUSIVE lock mode only.
> 
> The SELECT command acquires a lock of this mode on referenced tables. In general, any query that only reads a table and does not modify it will acquire this lock mode.
> 

So one query wants read access on TWO tables, while the other wants an exclusive access to the same TWO tables. One query grabs one lock first, the other query grabs the second one, and they both wait for the other table to become unlocked.

Postgres monitors this and when any process fails to obtain the lock it is waiting on for longer than 1 second, it will jump in and RANDOMLY kill one process in contention. This is also evident from the logs as failing queries have around 1s execution time.

<details>

<summary>Postgres logs describing the deadlock (many logs observed with the same footprint)</summary>

```
2026-01-08 09:26:22.327 UTC [2645] LOG:  00000: process 2645 detected deadlock while waiting for AccessExclusiveLock on relation 96369 of database 96312 after 1002.607 ms
2026-01-08 09:26:22.327 UTC [2645] DETAIL:  Processes holding the lock: 2646, 2645, 2640. Wait queue: .
2026-01-08 09:26:22.327 UTC [2645] CONTEXT:  SQL statement "CREATE TABLE IF NOT EXISTS gateway_envelope_blobs_o1200 PARTITION OF gateway_envelope_blobs FOR VALUES IN (1200) PARTITION BY RANGE (originator_sequence_id)"
	PL/pgSQL function make_blob_originator_part(integer) line 3 at EXECUTE
	SQL statement "SELECT make_blob_originator_part(p_originator_node_id)"
	PL/pgSQL function ensure_gateway_parts(integer,bigint,bigint) line 6 at PERFORM
2026-01-08 09:26:22.327 UTC [2645] LOCATION:  ProcSleep, proc.c:1494
2026-01-08 09:26:22.327 UTC [2645] STATEMENT:  -- name: EnsureGatewayParts :exec
	SELECT ensure_gateway_parts(
	               $1,
	               $2,
	               $3
	       )
	
2026-01-08 09:26:22.327 UTC [2645] ERROR:  40P01: deadlock detected
2026-01-08 09:26:22.327 UTC [2645] DETAIL:  Process 2645 waits for AccessExclusiveLock on relation 96369 of database 96312; blocked by process 2646.
	Process 2646 waits for AccessShareLock on relation 96358 of database 96312; blocked by process 2645.
	Process 2645: -- name: EnsureGatewayParts :exec
	SELECT ensure_gateway_parts(
	               $1,
	               $2,
	               $3
	       )
	
	Process 2646: -- name: SelectGatewayEnvelopesByOriginators :many
	WITH cursors AS (SELECT x.node_id AS cursor_node_id, y.seq_id AS cursor_sequence_id
	                 FROM unnest($1::INT[]) WITH ORDINALITY AS x(node_id, ord)
	                          JOIN unnest($2::BIGINT[]) WITH ORDINALITY AS y(seq_id, ord)
	                               USING (ord)),
	     filtered AS (SELECT m.originator_node_id,
	                         m.originator_sequence_id,
	                         m.gateway_time,
	                         m.topic
	                  FROM gateway_envelopes_meta AS m
	                           LEFT JOIN cursors AS c
	                                     ON m.originator_node_id = c.cursor_node_id
	                  WHERE m.originator_node_id = ANY ($3::INT[])
	                    AND m.originator_sequence_id > COALESCE(c.cursor_sequence_id, 0)
	                  ORDER BY m.gateway_time, m.originator_node_id, m.originator_sequence_id
	                  LIMIT NULLIF($4::INT, 0))
	SELECT f.originator_node_id,
	       f.originato
2026-01-08 09:26:22.327 UTC [2645] HINT:  See server log for query details.
2026-01-08 09:26:22.327 UTC [2645] CONTEXT:  SQL statement "CREATE TABLE IF NOT EXISTS gateway_envelope_blobs_o1200 PARTITION OF gateway_envelope_blobs FOR VALUES IN (1200) PARTITION BY RANGE (originator_sequence_id)"
	PL/pgSQL function make_blob_originator_part(integer) line 3 at EXECUTE
	SQL statement "SELECT make_blob_originator_part(p_originator_node_id)"
	PL/pgSQL function ensure_gateway_parts(integer,bigint,bigint) line 6 at PERFORM
2026-01-08 09:26:22.327 UTC [2645] LOCATION:  DeadLockReport, deadlock.c:1130
```
</details>

Postgres documentation related to table locks - [link](https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-TABLES).

Postgres documentation related to creating partition of a table - [link](https://www.postgresql.org/docs/current/sql-createtable.html)

## Change description

In the Postgres documentation for table partitioning, it is suggested that an alternative method could be used to create table partitions: a new table is created with insert constraints set in place as a standalone table, then an `ALTER TABLE Y ATTACH PARTITION X` is used to attach the partition. The constraint can then be dropped. `ATTACH PARTITION` query only requires an ShareUpdateExclusiveLock.

> As an alternative to creating a new partition, it is sometimes more convenient to create a new table separate from the partition structure and attach it as a partition later. [...] Moreover, the ATTACH PARTITION operation requires only a SHARE UPDATE EXCLUSIVE lock on the partitioned table rather than the ACCESS EXCLUSIVE lock required by CREATE TABLE ... PARTITION OF, so it is more friendly to concurrent operations on the partitioned table; [see [ALTER TABLE ... ATTACH PARTITION](https://www.postgresql.org/docs/18/sql-altertable.html#SQL-ALTERTABLE-ATTACH-PARTITION) for additional details. ]The [CREATE TABLE ... LIKE](https://www.postgresql.org/docs/18/sql-createtable.html#SQL-CREATETABLE-PARMS-LIKE) option can be helpful to avoid tediously repeating the parent table's definition.

This is the actual gist of the PR - essentially we change how we create partitions by creating the standalone tables and attaching them afterwards, which uses more relaxed locking which never collides with reads.

## Testing

Ran relevant tests in a loop thousands of times and deadlock issue did not appear. Before we hit the deadlock even with tens of repetitions.

One semi-harmless issue I did spot was a data race where we try to attach a table as a partition, we get `ERROR: "gateway_envelopes_meta_o100" is already a partition (SQLSTATE 42809)`. I believe it should be safe to ignore these errors, as it means another function beat us to it and created the meta/blobs tables, and we'd just need to adjust error handing for `ensure_gateway_parts` call.

### DIFF
--- a/pkg/migrations/00015_partition_management.up.sql
+++ b/pkg/migrations/00015_partition_management.up.sql
@@ -1,3 +1,8 @@
+-- NOTE: In case two processes try to attach the same table as a partition,
+-- Postgres will return a "wrong object type" error (SQLSTATE 42809) with a message
+-- 'table X is already a partition'. However, the error code 42809 is not exclusive
+-- to this scenario, so we check message text as a more precise indicator.
+
 -- META: create LIST child for one originator, then make it RANGE-partitioned
 CREATE OR REPLACE FUNCTION make_meta_originator_part_v2(_oid int)
     RETURNS void AS $$
@@ -31,7 +36,14 @@ BEGIN
         ALTER TABLE %I DROP CONSTRAINT oid_check;',
         subname
     );
-
+EXCEPTION
+    WHEN OTHERS THEN
+        IF SQLERRM ~ 'is already a partition' THEN
+            -- Do nothing.
+            NULL;
+        ELSE
+            RAISE;
+        END IF;
 END
 $$ LANGUAGE plpgsql;
 
@@ -70,7 +82,14 @@ BEGIN
         'ALTER TABLE %I DROP CONSTRAINT seq_id_check;',
         subname
     );
-
+EXCEPTION
+    WHEN OTHERS THEN
+        IF SQLERRM ~ 'is already a partition' THEN
+            -- Do nothing.
+            NULL;
+        ELSE
+            RAISE;
+        END IF;
 END;
 $$ LANGUAGE plpgsql;
 
@@ -108,7 +127,14 @@ BEGIN
         ALTER TABLE %I DROP CONSTRAINT oid_check;',
         subname
     );
-
+EXCEPTION
+    WHEN OTHERS THEN
+        IF SQLERRM ~ 'is already a partition' THEN
+            -- Do nothing.
+            NULL;
+        ELSE
+            RAISE;
+        END IF;
 END;
 $$ LANGUAGE plpgsql;
 
@@ -148,7 +174,14 @@ BEGIN
         'ALTER TABLE %I DROP CONSTRAINT seq_id_check;',
         subname
     );
-
+EXCEPTION
+    WHEN OTHERS THEN
+        IF SQLERRM ~ 'is already a partition' THEN
+            -- Do nothing.
+            NULL;
+        ELSE
+            RAISE;
+        END IF;
 END;
 $$ LANGUAGE plpgsql;
 


mkysel (changes requested):
you can not change an existing migration. A numbered migration is only executed once. So servers that have already executed 00002 will never do so again and your changes will never take effect.

You must create a new migration with new DDL.

PR #1596, opened 2026-02-03
Trigger for updating latest envelope runs per statement

Maelkum:
Instead of running the trigger per row like before, new trigger runs per statement, which should be more performant on batch inserts.

### Alternative Implementation

In the `update_latest_envelope_v2` function, the inner window function could be replaced with something like this:

```sql
SELECT originator_node_id, MAX(originator_sequence_id) AS originator_sequence_id, MAX(gateway_time) AS gateway_time
FROM new
GROUP BY originator_node_id
```

This could work, but I'm not sure if there is potentially a chance that one row has a higher sequence ID, and another row higher gateway time? With a query like this we might take max values that actually come from two different rows. The `ROW_NUMBER` version will force the gateway_time of the row with the largest sequence ID.

### DIFF
--- a/pkg/migrations/00018_add_latest_envelopes.down.sql
+++ b/pkg/migrations/00018_add_latest_envelopes.down.sql
@@ -1,2 +1,9 @@
+-- First reinstate previous version of the trigger, as defined in 00006_add_latest_envelopes.up.sql
+CREATE TRIGGER gateway_latest_upd
+    AFTER INSERT ON gateway_envelopes_meta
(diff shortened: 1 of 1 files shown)


Maelkum:
I was aware that `gateway_time` is of lesser importance - I consider it more as a descriptive thing; however I figured it might be slightly confusing if the two "max" values come from different rows. But, since gateway time is not that relevant, I figure we go with the simpler option which should be faster too.

The suggestion from Macroscope makes sense though - should I reinstate previous version of trigger on `down` migration?

I don't think we have such a case in the project yet.

PR #1605, opened 2026-02-05
Insert and revoke address logs in batches

fbac:


neekolas (approved):
I'll approve. I am not sure how effective this is going to be in practice.

I would imagine something like 99.9% of address log updates are 1:1 with identity updates. The protocol supports batch operations, but that was more of a future-proofing decision than anything else.

PR #1608, opened 2026-02-06
Introduce database worker to check and create partitions

Maelkum:
## Description

This PR introduces a background database worker that will periodically check existing partitions and create new ones if the existing partitions are sufficiently full. In the future this database worker can be extended to run more background tasks, as needed.

Worker starts a background loop that runs a check every 30 minutes. We query `nformation_schema.tables` Postgres table to get the list of existing tables, using the regex to match our range partitions for gateway envelopes (e.g. `gateway_envelopes_meta_o100_s0_1000000`). For these partitions, per originator ID, we identify the _last_ partition, and check how full it is. To determine how full it is we use the last sequence ID in that table/partition, and how close it is to the partition upper range. For example, given a partition size of 10 with a single envelope with sequence ID of 7 is considered 70% full - as the subsequent envelopes will likely be 8, 9 and 10, and not 1, 2 or 3.

For a full partition we will then create the next one, with a configurable partition size. Current code continues with the chosen bandwidth of 1M for next partitions.

Failure to run checks and create partitions is logged but does not halt the node, as there's always fallback for partition creation in our gateway insertion functions.

I used sqlc to create queries in a separate package compared to our main `db` one as to not pollute the namespace with less common, utility functions. I hate that I can't configure sqlc a bit more to not emit even some of the methods it emits right now, so there's some stuff that we don't need.

Right now from outside the package the worker can only be created with default configuration - it didn't seem too necessary for me to provide too much external configuration at this moment, but it can be added if needed.

### DIFF
--- a/pkg/db/worker/worker.go
+++ b/pkg/db/worker/worker.go
@@ -33,11 +33,7 @@ type Config struct {
 }
 
 // partition config controls when should the database worker create new partitions.
-// given the partition size, it will create new partitions when the the partition
-// is fill more than the specified fill threshold.
-// For example: partition size is 1000, fill threshold is 70% => when the partition
-// has over 700 entries it will create the next partition.
-// TODO: Perhaps use a list of originators and setup partitions even for nodes that do not have any yet.
+// TODO: Perhaps use a lis of originators and setup partitions even for nodes that do not have any yet.
 
 type PartitionConfig struct {
 	FillThreshold float64


pkg/db/worker/worker.go
```diff
@@ -0,0 +1,223 @@
...
+
+	"go.uber.org/zap"
+
+	"github.com/xmtp/xmtpd/pkg/db"
+	"github.com/xmtp/xmtpd/pkg/db/queries"
+	"github.com/xmtp/xmtpd/pkg/utils"
+)
+
+const (
+	DefaultPartitionSize = 1_000_000
```
fbac: Since the bandwidth is going to be used by different packages I'd advice on moving the constant to pkg/constants and modifying everything to use it. We don't want different sources of truth.

fbac (commented):
Made a first review pass and left some comments and doubts.

pkg/db/worker/worker.go
Maelkum: Right, will do!

mkysel (changes requested):
is there a good reason for duplication?

PR #1619, opened 2026-02-11
Test deterministic db migration output

fbac:


fbac:
> LGTM!
> 
> Curious - why mixing of asserts alongside requires for checking?

I don't have a technical answer to that, I use them interchangeably. Mostly I use require for Error or NotError automatically, so sometimes I mix them.

If there are technical considerations on how to use them happy to learn!

Issue #1694, opened 2026-02-20
Fix sync service


fbac:
The sync service broke because of a subtle bug introduced in the renewed SelectGatewayEnvelopeByOriginators.

With more than one originator ID it can happen that:
- service tries to subscribe to 100, 200, 300
- the limit imposed per originator is max_envelopes/num_originators
- the condition `if receivedEnvs < maxEnv { break }` is hit.

PR #1695, opened 2026-02-20
Fix sync service

fbac:
This PR reverts the SelectGatewayEnvelopesByOriginators to its previous state.

Introduces a regression test which makes sure that:
- Syncing from originators will always catch up completely. That's why 1001 envelopes are used, to test also the max rows per request (1000).
- Syncing multiple originator ids won't break the system invariant that all messages from all those originator ids are synced.


fbac:
Closes #1694

PR #1983, opened 2026-04-17
fix: accept unknown originators in SubscribeTopics cursor

### LINKED ISSUE #1982: SubscribeTopics fails when migrator originator included in cursor in environments without migrator
## Problem

`libxmtp` can send requests using a migrator `originator_id` (10 or 11) even in environments where the migrator has never been configured. I think this may be triggered by optimistic message sends, but need to confirm.

In `SubscribeTopics` we validate that all originators included in a cursor actually exist on the network and will reject subscription requests that include cursors referencing for unknown originators.

The node's originator list is populated by the `gateway_envelopes_latest` table, and requires the node to have seen at least 1 message from that originator.

## Proposed solution

Relax validation of provided cursors and allow unknown originators to be specified. Unknown originators are not that harmful (clients just won't get results) and this may help prevent other complex bugs around network partitions or very new nodes. It may be possible for a client to know of a new originator before a node does, or for the client to have messages from a node that was removed long ago and newer nodes have received 0 messages from that originator.

xmtp-coder-agent:
Resolves https://github.com/xmtp/xmtpd/issues/1982

## Summary

`SubscribeTopics` rejected requests with `CodeInvalidArgument` whenever a
client's per-topic cursor referenced an originator the node had not yet
indexed (i.e. absent from `gateway_envelopes_latest`). This breaks
legitimate flows:

- `libxmtp` sends cursors containing migrator originator IDs (10, 11) even
  in environments where no migrator has been configured.
- A client may learn of a new originator before a given node has indexed
  any of its messages (new nodes, network partitions).
- A client may still hold a cursor for an originator that was removed long
  ago and has no messages on any currently-live node.

The originator set is node-local and lossy — treating a mismatch as a fatal
client error is the wrong default.

## Change

- Remove the unknown-originator rejection loop in `validateTopicFilters`.
  All other validation (filter count, topic length, vector clock length)
  stays in place.
- `buildTopicCursors` already tolerated unknown originators: the client's
  entries pass through unchanged, and `FillMissingOriginators` still seeds
  known originators at sequence 0. The downstream
  `SelectGatewayEnvelopesByPerTopicCursors` is a `CROSS JOIN LATERAL` on
  the client's `(topic, node_id, seq_id)` tuples — entries that don't
  match any row contribute nothing.
- `advanceTopicCursors` already handles previously-unknown originators
  starting to publish live, so that path needed no change.

## Test plan

- [x] Replaced `TestSubscribeTopics_UnknownOriginatorInCursor` (previously
      asserted rejection) with `TestSubscribeTopics_AcceptsUnknownOriginatorInCursor`,
      which asserts the subscription opens and delivers envelopes from
      known originators.
- [x] Added `TestSubscribeTopics_MixedKnownAndUnknownOriginators`: cursor
      mixes a partially caught-up known originator with an unknown one;
      catch-up delivers only the unseen messages for the known originator.
- [x] Added `TestSubscribeTopics_AllUnknownOriginatorsInCursor`: all cursor
      entries are unknown; catch-up still delivers messages from known
      originators via `FillMissingOriginators`.
- [x] `TestSubscribeTopics_Validation` unchanged — other validation errors
      still return `CodeInvalidArgument`.
- [x] `go test ./pkg/api/message/... -count=1`
- [x] `dev/lint-fix`

> [!NOTE]
> ### Accept unknown originator IDs in `SubscribeTopics` cursor without error
> Previously, `validateTopicFilters` rejected cursors containing originator IDs not known to the node. This removes that check, since unknown originators simply match no rows downstream and should not block subscription processing.
>
> - Removes the `knownOriginators` parameter from [`validateTopicFilters`](https://github.com/xmtp/xmtpd/pull/1983/files#diff-ac39f1802be2ab7789d5ffc0d2344a0d6f82d8b6a85fde0dc434a0069097bcff), dropping the loop that rejected unknown originator IDs in `LastSeen` cursors.
> - Moves `validateTopicFilters` to run before the known-originator lookup, simplifying the handler flow.
> - Behavioral Change: cursors with unknown originator IDs are now accepted; catch-up queries for those originators return no results rather than an error.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 426f102.</sup>
> 
>

### DIFF
--- a/pkg/api/message/subscribe_topics.go
+++ b/pkg/api/message/subscribe_topics.go
@@ -68,16 +68,16 @@ func (s *Service) SubscribeTopics(
 		)
 	}
 
+	if err := validateTopicFilters(filters); err != nil {
+		return err
+	}
+
 	knownOriginators, err := s.originatorList.GetOriginatorNodeIDs(ctx)
 	if err != nil {
 		return connect.NewError(
 			connect.CodeInternal,
 			fmt.Errorf("could not get originator list: %w", err),
 		)
-	}
-
-	if err := validateTopicFilters(filters, knownOriginators); err != nil {
-		return err
 	}
 
 	cursors, topics, topicKeys := buildTopicCursors(filters, knownOriginators)
@@ -146,9 +146,13 @@ func (s *Service) SubscribeTopics(
 }
 
 // validateTopicFilters validates the topic filters in a SubscribeTopicsRequest.
+// Cursor entries for originators the node has not yet seen are allowed: a
+// client may learn of a new originator before this node has indexed any of
+// its messages, or may still hold cursors for originators removed long ago.
+// Unknown originators are harmless in the downstream LATERAL query — they
+// simply match no rows.
 func validateTopicFilters(
 	filters []*message_api.SubscribeTopicsRequest_TopicFilter,
-	knownOriginators []uint32,
 ) error {
 	if len(filters) == 0 {
 		return connect.NewError(
@@ -164,23 +168,9 @@ func validateTopicFilters(
 		)
 	}
 
-	known := make(map[uint32]struct{}, len(knownOriginators))
-	for _, id := range knownOriginators {
-		known[id] = struct{}{}
-	}
-
 	for _, f := range filters {
 		if err := validateTopicFilter(f); err != nil {
 			return connect.NewError(connect.CodeInvalidArgument, err)
-		}
-
-		for origID := range f.GetLastSeen().GetNodeIdToSequenceId() {
-			if _, ok := known[origID]; !ok {
-				return connect.NewError(
-					connect.CodeInvalidArgument,
-					fmt.Errorf("unknown originator node ID in cursor: %d", origID),
-				)
-			}
 		}
 	}
 
--- a/pkg/api/message/subscribe_topics_test.go
+++ b/pkg/api/message/subscribe_topics_test.go
@@ -189,7 +189,7 @@ func TestSubscribeTopics_Validation(t *testing.T) {
 	}
 }
 
-func TestSubscribeTopics_UnknownOriginatorInCursor(t *testing.T) {
+func TestSubscribeTopics_AcceptsUnknownOriginatorInCursor(t *testing.T) {
 	client, store, _ := setupTopicTest(t)
 	payerID := db.NullInt32(testutils.CreatePayer(t, store))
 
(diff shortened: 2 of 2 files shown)


xmtp-coder-agent:
CI failure is unrelated to this PR. The failing test `TestCreateServer` in `pkg/server/server_test.go:163` is a known cross-node replication flake (previously stabilized in #1975). It timed out on a `require.Eventually` waiting for envelopes to replicate between two test servers — this PR does not touch any publish-worker, sync, or cross-node replication paths.

Verified locally:
```
go test ./pkg/server/... -run 'TestCreateServer$' -count=1
ok  	github.com/xmtp/xmtpd/pkg/server	13.242s
```

I lack admin rights to re-run the failed job; requesting a re-run from a maintainer.

xmtp-coder-agent:
Reviewed each Octane finding against the actual change. None require action in this PR — details below.

**High: "Unknown-originator cursor acceptance ... causes resource-exhaustion DoS"** — not valid.
Per-subscription resource ceilings are unchanged:
- `maxTopicFilters = 10_000` filters per request (`subscribe_topics.go:30`).
- `maxVectorClockLength = 100` entries per filter (`service.go:45`, enforced in `validateTopicFilter`).
- `applySubscribeAdmission` already charges `ceil(sqrt(numFilters))` tokens per stream open.

Before this PR, a client could already send the maximum-sized cursor using only known originator IDs. Allowing the same size cursor to reference unknown IDs does not raise the ceiling on memory, CPU, or DB work. The LATERAL probe for a never-seen originator terminates on the index immediately (no matching rows). The global concurrent-stream limit is a separate, pre-existing concern tracked in #1957.

**Medium: "Relaxed cursor validation ... during mixed-version rollout causes intermittent subscription failures"** — inverts the direction.
This PR makes the server *more* tolerant, not less. During a partial rollout, unpatched nodes continue rejecting unknown-originator cursors (today's behavior); patched nodes accept them. User-observable failure rate monotonically decreases as the fix rolls out — no new failure mode is introduced.

**Low: "Missing eviction of invalid nodes in SmartContractRegistry"** in `pkg/registry/node_registry_contract.go` — out of scope.
This file is not modified in this PR.

**Warning: "Relaxed validation combined with first-wins duplicate-topic deduplication"** — pre-existing.
The first-wins dedup at `buildTopicCursors` (subscribe_topics.go:220) is unchanged by this PR. If a client sends duplicate topic filters with divergent cursors, the first still wins — same as before.

**Warning: "uint64/int64 per-topic cursor mismatch"** — pre-existing.
`VectorClock` is `map[uint32]uint64`; the SQL params are `int64`. Both predate this PR and are unchanged. This PR only modifies `validateTopicFilters` — no cursor-type handling was touched.

Happy to address any of these in a follow-up if a maintainer considers one legitimate; leaving this PR focused on the scope of #1982.

fbac (approved):

PR #1917, opened 2026-04-02
feat: register new d14n gRPC services and implement SubscribeOriginators

neekolas:
## Summary

Phase 1 of the d14n service restructuring ([xmtp/proto#327](https://github.com/xmtp/proto/issues/327)). Registers the new gRPC services introduced by [xmtp/proto#328](https://github.com/xmtp/proto/pull/328) on the xmtpd server and implements the only new RPC method, `SubscribeOriginators`.

### New services registered

| Service | Server | Backed by | Methods |
|---------|--------|-----------|---------|
| `QueryApi` | Replication | `message.Service` | `QueryEnvelopes`, `SubscribeTopics`, `GetInboxIds`, `GetNewestEnvelope` |
| `PublishApi` | Replication | `message.Service` | `PublishPayerEnvelopes` |
| `NotificationApi` | Replication | `message.Service` | `SubscribeAllEnvelopes` |
| `GatewayApi` | Gateway | `payer.Service` | `PublishClientEnvelopes`, `GetNodes` |

No new structs — the existing service structs satisfy the new interfaces because the proto changes use identical method signatures and request/response types.

### New RPC: `SubscribeOriginators`

Added to `ReplicationApi`. Subscribes to envelopes from specific originator nodes with cursor-based catch-up, using the same `subscribeWorker` infrastructure as `SubscribeEnvelopes`.

- `LastSeen` cursor is **required** (nil returns `InvalidArgument`)
- Responses use the `Envelopes` oneof variant, leaving room for future status signals
- Envelope batching accounts for the oneof wrapper overhead to avoid exceeding gRPC payload limits

### Internal type: `subscribeFilter`

Replaced internal usage of the deprecated `message_api.EnvelopesQuery` proto with `subscribeFilter`, an unexported type with explicit `catchUpMode` enum:

```go
type catchUpMode int
const (
    catchUpNone      catchUpMode = iota // stream new envelopes only
    catchUpFromStart                     // catch up from the very beginning
    catchUpFromCursor                    // catch up from specified cursor position
)
```

Each subscribe endpoint sets the mode explicitly at its entry point — no implicit nil cursor semantics.

### Shared subscribe helpers

Extracted `batchAndSendEnvelopes` and `catchUpWithSendFn` into `subscribe_common.go`, eliminating ~200 lines of duplication between `SubscribeEnvelopes` and `SubscribeOriginators`. Both use callback-based response wrapping so the batching, cursor dedup, and pagination logic is written once.

### Deprecated endpoints

[...]

PR #1989, opened 2026-04-22
feat(sync): migrate sync worker to SubscribeOriginators with fallback

### LINKED ISSUE #1980: sync: migrate sync worker from SubscribeEnvelopes to SubscribeOriginators
## Context

PR #1917 introduced a new `ReplicationApi.SubscribeOriginators` RPC
(`pkg/api/message/subscribe_originators.go`) as part of the d14n service
split. It replaces the node-to-node sync use case of the older
`SubscribeEnvelopes` RPC with a narrower, stricter contract:

- Request shape is a flat `OriginatorFilter{originator_node_ids, last_seen}`
  instead of the nested `EnvelopesQuery`.
- Requires ≥1 originator ID and a non-nil `last_seen` cursor
  (enforced in `validateOriginatorFilter`).
- Drops topic filtering, which the sync path never used.

The sync worker still calls `SubscribeEnvelopes`, so the new endpoint is
unused on the consumer side.

## Goal

Switch the sync worker to call `SubscribeOriginators` so that node-to-node
replication uses the dedicated RPC and the legacy `EnvelopesQuery` path
can eventually be retired from the sync code.

## Scope

- `pkg/sync/sync_worker.go` — swap the RPC call and request construction
  (`EnvelopesQuery` → `OriginatorFilter`) around the current call site at
  `setupStream`.
- `pkg/sync/originator_stream.go` — update the stream client type and
  `Recv()` response unwrap to `SubscribeOriginatorsResponse`.
- Tests under `pkg/sync/` — update fakes/mocks for the new stream type.

## Out of scope

- No changes to cursor/vector-clock semantics, dedup, OOO handling,
  migrator-originator allowlist, or `envelope_sink`.
- No changes to `SubscribeEnvelopes` itself (other consumers still use it).
- Network-watcher is unaffected (it uses `MetadataApi.SubscribeSyncCursor`).

## Acceptance

- Sync worker establishes streams via `SubscribeOriginators` against peers.
- Existing sync tests pass with the new RPC, including migration-mode
  coverage (migrator originator IDs in the filter).
- No regression in vector-clock progression or OOO/gap detection.

mkysel:
## Summary

Closes #1980. Migrates the sync worker from `ReplicationApi.SubscribeEnvelopes` to the new `ReplicationApi.SubscribeOriginators` RPC introduced by #1917. To keep node-to-node sync working during rollout, `SubscribeEnvelopes` is used as a transparent fallback when a peer returns `codes.Unimplemented`.

- Introduces an `envelopeRecvStream` interface with two adapters (`originatorsStreamAdapter`, `envelopesStreamAdapter`) so the existing reader loop stays RPC-agnostic. Keepalive frames on the new RPC return an empty slice and are skipped.
- Adds a `subscribeWithFallback` helper that prefers `SubscribeOriginators` and only falls back on `codes.Unimplemented`. The decision is evaluated per stream setup — no per-peer caching — so peers upgraded between reconnects are picked up automatically.
- Adds `xmtp_sync_subscribe_rpc_total{rpc, originator_id}` counter to observe the rollout. Emits an Info log on every fallback with the peer node ID and HTTP address.
- No changes to vector-clock / dedup / OOO-detection / migrator-allowlist / envelope-sink behavior.

## Test Plan

- [x] `go test ./pkg/sync/... -race -count=1` — all tests pass, including migration-mode and OOO coverage
- [x] New unit tests: envelope/originators adapters (incl. keepalive), and all four branches of `subscribeWithFallback` (success, fallback, non-`Unimplemented` propagation, envelopes-branch error propagation)
- [x] `dev/lint-fix` clean
- [x] `go build ./...` clean
- [ ] Manual verification against a v1.2.0 peer to confirm fallback log + metric fire
- [ ] Manual verification against a current-version peer to confirm `SubscribeOriginators` is used

## Rollout removal

Once all registered nodes report a version shipping `SubscribeOriginators`, delete the fallback branch, the `envelopes` adapter, and the `rpc="envelopes"` metric label value. Track as a follow-up issue.

> [!NOTE]
> ### migrate sync worker stream setup to prefer `subscribeOriginators` with `subscribeEnvelopes` fallback
> - `syncWorker.setupStream` now calls `subscribeWithFallback`, which probes `subscribeOriginators` first by consuming the initial frame; if the server returns `codes.unimplemented`, it falls back to `subscribeEnvelopes`.
> - A new `envelopeRecvStream` interface abstracts both RPC streams, with `envelopesStreamAdapter` and `originatorsStreamAdapter` implementing it; the originator adapter handles keepalive (empty frames) and first-frame replay.
> - A new `xmtp_sync_subscribe_rpc_total` counter metric (labels: `rpc`, `originator_id`) records which RPC path was taken for each stream setup.
> - Risk: the fallback probe eagerly consumes the first frame from `subscribeOriginators`; if the frame is lost before replay, envelopes in that frame could be missed on non-adapter code paths.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5de1338.</sup>
> 
>


mkysel:
Pushed two fixes in response to the Octane findings.

**Addressed**

- **#2 MEDIUM — missing collector registration.** Added `syncSubscribeRPC` to `registerCollectors()` in `pkg/metrics/metrics.go`. Without this the counter was incremented but never scraped. (`29e4f1b0`)
- **#3 LOW — setup-time-only capability check.** For server-streaming gRPC, `codes.Unimplemented` comes back in trailers on the first `Recv()`, not from the client method itself. The old fallback only checked the method's error, so against a real un-upgraded peer the `Unimplemented` would have slipped past the fallback branch and the stream listener would have retried forever. `subscribeWithFallback` now probes via the first `Recv()` (the server sends an immediate keepalive so this unblocks promptly), detects `Unimplemented` from either path, and stashes the probed frame on the adapter so the reader loop replays it without loss. Tests updated to model real gRPC semantics plus a defensive case for method-level Unimplemented. (`5de13389`)

**Not addressed (out of scope)**

- **#1 HIGH — 10-byte wrapper overhead / cursor-advance on oversized envelopes in `subscribe_common.go`.** Pre-existing server code from #1917; not touched by this PR. Worth its own issue/PR.
- **#4 INFO — missing per-call timeout on the probe.** Sync streams are long-lived and already gated by the sync worker's context + outer backoff retry loop. Adding a setup-only deadline risks breaking steady-state operation without a clear problem to solve.
- **Warning — single-envelope limit in `service.go`.** Pre-existing server code from #1917; not touched by this PR.

neekolas (approved):
Do we actually need a fallback, if we push this out as part of the next major release?

mkysel:
> Do we actually need a fallback, if we push this out as part of the next major release?

Not strictly necessary but it was easy and it is the right way to do things. If it were extra work, I would have chosen a hacky approach.

PR #2026, opened 2026-06-23
fix(db): eliminate gateway-envelope partition-creation deadlock via reader/writer advisory lock

tylerhawkes:
## Problem

`TestCreateServer` was failing intermittently, and the root cause turned out to be a **real production availability bug**, not a test-only flake.

Lazy, on-the-fly partition creation deadlocks under concurrent writers. When a gateway-envelope insert hits a missing partition, `ensure_gateway_parts_v3` runs `ALTER TABLE … ATTACH PARTITION` against the two shared parent tables `gateway_envelopes_meta` and `gateway_envelopes_blob`.

The lock-ordering inversion (proven with `pg_locks` + the actual deadlock reports):

- `gateway_envelopes_blob` has an FK → `gateway_envelopes_meta`, and `gateway_envelopes_meta` has an FK → `payers`.
- Once the parents hold data, each `ATTACH` validates the inherited FK and takes `ShareRowExclusiveLock` (SROX) on **both** parents. SROX conflicts with the `RowExclusiveLock` an ordinary insert takes.
- A single transaction therefore touches **meta → blob → meta** (the last via the blob insert's FK check). Two concurrent transactions for **different originators** — e.g. the publish-worker for its own originator and a sync-worker replicating another — acquire the two parents' locks in opposite orders → deadlock (SQLSTATE 40P01).

The publish-worker exhausts its 3 retries and then sleeps, so it can stall ingest. The test amplifies it: fresh DBs mean both originators create their partitions at startup simultaneously.

**Production impact:** every node runs a publish-worker plus N sync-workers, all writing to one DB for different originators. Partition creation fires whenever a node first sees a new originator, or when any originator crosses a 1,000,000-sequence band boundary. Two such events concurrently hit the exact inversion. Rare, but real.

## Fix: reader/writer advisory lock + out-of-band partition creation

- Ordinary gateway inserts take the **shared** side of a new transaction-scoped advisory lock (`LockKindPartitionCreation`), so they still run concurrently with each other.
- Partition creation runs in its **own** transaction under the **exclusive** side (`EnsureGatewayPartitions`), so DDL never overlaps DML — removing the SROX-vs-`RowExclusive` conflict entirely. It is a separate transaction specifically to avoid a shared→exclusive upgrade deadlock.
- Insert helpers now return a typed `ErrGatewayPartitionMissing` instead of creating partitions inline; the transaction-owning callers (`publish_worker`, `migrator`, `payerreport`, the identity-update indexer) create the partition out-of-band and retry. The publish path takes the shared lock before resolving payers, for a consistent lock order.

## Verification

- **Deadlock eliminated:** clean-start comparison, 50 `-race` iterations each — **baseline `main`: 17 Postgres deadlocks; this branch: 0.** Across 300+ total `-race` `TestCreateServer` iterations: **0** deadlock detections, **0** publish-worker batch retries.
- `golangci-lint` v2.10.1, `gofmt`, `go vet`, `go build ./...`: clean.
- Unit tests pass: `pkg/db` (race, 2×), `pkg/migrator`, `pkg/payerreport`, `pkg/api/message`, `pkg/indexer/app_chain/contracts`.

## A separate, pre-existing flake remains (out of scope here)

`TestCreateServer` also has an **independent replication flake** — the envelope publishes fine but cross-node replication stalls (`Condition never satisfied`). That is issue #1970 (previously patched in #1975), unrelated to this deadlock and **not** introduced here: a clean-start comparison shows the same ~4% rate on both baseline and this branch. It should be tracked and fixed separately.

## Diff note

The large diffs in `identity_update_storer.go` and `payerreport/store.go` are mostly closure-extraction reindentation (wrapping the insert in the ensure-and-retry contract); the actual logic change in each is small.

> [!NOTE]
> ### Fix gateway-envelope partition-creation deadlock using reader/writer advisory locks
> - Replaces inline DDL partition creation inside insert transactions with an out-of-band ensure-and-retry pattern across all gateway envelope insert paths ([gateway_envelope.go](https://github.com/xmtp/xmtpd/pull/2026/files#diff-a41762f1583f16a22a90bd90a773d5e56ab4d77754ba991347c8a67b8e1229db), [gateway_envelope_batch.go](https://github.com/xmtp/xmtpd/pull/2026/files#diff-3b81b47f7dc103297421e66828594182df1ae56fb06cd67801cdd424621be120), [publish_worker.go](https://github.com/xmtp/xmtpd/pull/2026/files#diff-a985e2d18aa8bb7ea586f77e5bc52dad0a4c4989e5a220806f737fb1c9c50415), [identity_update_storer.go](https://github.com/xmtp/xmtpd/pull/2026/files#diff-719fcd3f6b2a547b6c6f299d70464725ddf9c1161c104099b3c3cab140cb2354), [migrator/writer.go](https://github.com/xmtp/xmtpd/pull/2026/files#diff-79de0174d18bbb97eb4c62be1eeec1e2132ff3060f5beafb3b8721044791b258), [payerreport/store.go](https://github.com/xmtp/xmtpd/pull/2026/files#diff-1557531d1361293f11123bbcc86167d7a99f261d62a4a71af45f898ac3b5c770)).
> - Transactional inserts now acquire a shared (reader) advisory lock via `SharedLockPartitionCreation`; partition creation acquires an exclusive (writer) lock via `LockPartitionCreation`, serializing DDL against concurrent DML.
> - When an insert fails due to a missing partition, the transactional function returns `ErrGatewayPartitionMissing` instead of creating the partition inline; callers then call `EnsureGatewayPartitions`/`EnsureGatewayPartitionsForBatch` outside the transaction and retry once.
> - Risk: all callers of transactional gateway insert functions must now handle `ErrGatewayPartitionMissing` and perform the ensure-and-retry; callers that do not are updated in this PR but any future callers must follow the same contract.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 4f4aeb2. 7 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/pkg/api/message/publish_worker.go
+++ b/pkg/api/message/publish_worker.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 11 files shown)


mkysel (approved):
good find. this has been flaring up every now and then and we never prioritized looking into it.

PR #2057, opened 2026-08-26
fix(indexer): stop probing the next header at chain tip

### LINKED ISSUE #1879: Unconditional next-block header fetch at chain tip in RPCLogStreamer.GetNextPage causes backfill stall and noisy retries
**Severity:** Low | **Likelihood:** Low | **Impact:** Medium | **Type:** Vulnerability

## Details

GetNextPage performs a reorg check by fetching header(fromBlock+1) before verifying fromBlock < highestBlock; at chain tip this header does not exist, causing repeated errors and a stall until a new block appears.

In RPCLogStreamer.GetNextPage, the function calls HeaderByNumber(fromBlockNumber+1) for reorg detection whenever fromBlockNumber > 0, before ensuring that fromBlockNumber is below the current head. At the chain tip (fromBlockNumber == highestBlock), the next block header is not yet available, so HeaderByNumber(fromBlock+1) fails. The caller (watchContract) treats this as a generic error, sleeps 100ms, and retries, never reaching ErrEndOfBackfill or switching to subscription mode. This causes repeated error logs and a temporary stall at the backfill boundary.

### Scenario 1
Malicious or degraded RPC provider returns errors for HeaderByNumber(fromBlock+1) around the backfill cursor, preventing GetNextPage from reaching end-of-backfill and keeping the indexer stuck retrying indefinitely without switching to subscription.

**Preconditions / Assumptions:**
- (a) Node uses an untrusted or malfunctioning RPC provider for rpcClient
- (b) Indexer is near or at the chain tip (fromBlockNumber >= highestBlock)
- (c) Provider persistently returns errors to HeaderByNumber(fromBlock+1)

### Scenario 2
On node restart near chain tip with a quiet chain, GetNextPage probes header(fromBlock+1) which does not exist yet, returning an error; the backfill loop logs, sleeps, and retries until a new block is mined, stalling indexing and producing noisy logs.

**Preconditions / Assumptions:**
- (a) Node previously indexed up to chain head for the watched contracts
- (b) Operator restarts the node
- (c) Chain is idle or slow so the next block is not yet available when GetNextPage runs
- (d) RPC provider returns NotFound for non-existent headers

### Scenario 3
While the node is stalled at the backfill boundary, an attacker emits many matching events to the watched contract, filling the subscription channel; this causes backpressure and potential subscription reconnection churn until backfill eventually completes.

**Preconditions / Assumptions:**
- (a) Node is already stalled at the end-of-backfill boundary
- (b) Watched contracts are publicly callable
- (c) Attacker is willing to pay gas to emit many matching logs during the stall

## Files impacted

- `pkg/indexer/rpc_streamer/rpc_log_streamer.go`

**Lines 355-359:**
```go
// Do not check for reorgs at block height 0. Genesis does not have a parent.
if fromBlockNumber > 0 {
    nextBlockHeader, err := r.rpcClient.HeaderByNumber(
        ctx,
```

huklaa:
Fixes #1879.

## Problem
`GetNextPage` performed its reorg check by fetching `HeaderByNumber(fromBlock+1)` before it knew whether `fromBlock` was already the current head. At the head, `H+1` does not exist yet, so providers can return not-found and the backfill loop retries instead of fetching block H and transitioning to subscription mode.

The first fix avoided the nonexistent H+1 probe, but that also exposed a boundary gap: a 32-byte stored cursor hash was no longer validated at the chain tip.

## Fix
- validate `fromBlockNumber` against the reported head before the reorg probe;
- below the head, keep validating H through `H+1.ParentHash`;
- at the head, validate the stored cursor directly against `HeaderByNumber(H).Hash()` instead of probing nonexistent H+1;
- on mismatch, preserve the existing one-block rollback / `ErrReorg` behavior;
- when the cursor is canonical, still fetch the current head block and return `ErrEndOfBackfill` as before.

## Regression coverage
- `TestGetNextPageAtChainTipSkipsNextHeaderProbe` verifies a canonical tip cursor fetches H, returns its logs with `ErrEndOfBackfill`, and never probes H+1.
- `TestGetNextPageAtChainTipDetectsCursorReorg` verifies a stale tip cursor is detected before log fetching and rolls back one block, without probing H+1.

Local execution is unavailable in this automation environment; hosted PR CI is the execution gate for the focused tests.

> [!NOTE]
> ### Stop probing next block header at chain tip in `RPCLogStreamer.GetNextPage`
> - When at the chain tip (`fromBlockNumber == highestBlock`), validates the cursor directly against `header(H)` instead of probing the non-existent `header(H+1)`.
> - Returns an early error if `fromBlockNumber` exceeds `highestBlock`.
> - Reorders metrics to emit `EmitIndexerMaxBlock` immediately after fetching `highestBlock` and `EmitIndexerCurrentBlockLag` after validation.
> - Behavioral Change: `RPCLogStreamer.GetNextPage` now detects reorgs at the chain tip and returns `ErrReorg` stepping back to `H-1`, rather than failing on a future-header fetch.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 268db5f.</sup>
> 
>


huklaa:
Reviewed the Octane findings against the actual `GetNextPage` control flow and the new regression test.

- The “missed head-block logs” finding does not match the patch: at `fromBlock == highestBlock`, the function still calls `FilterLogs(H..H)` and the regression test asserts those logs are returned with `ErrEndOfBackfill`.
- There is, however, a real reorg-validation gap worth tightening: below the head we validate the stored cursor hash via `H+1.ParentHash`; at the head, skipping `H+1` also skips validation of a 32-byte stored hash entirely. The safer shape is to validate `HeaderByNumber(H).Hash()` against `fromBlockHash` when `fromBlock == highestBlock`, while keeping the existing `H+1.ParentHash` check for `fromBlock < highestBlock`. That avoids probing a nonexistent future header without weakening canonical-chain verification at the boundary.

I’m treating that as the substantive follow-up from the scan; the unrelated `publish_worker.go` finding is outside this PR’s diff/scope.

PR #2060, opened 2026-09-01
fix(stress): stop watchers when channels close

colinaumaty:
Return when subscription or log channels close instead of processing zero values. This prevents repeated empty events and lets the watcher goroutine terminate.

No protocol behavior changes; this fixes stress watcher lifecycle only.

## Testing

Run the stress package tests with PostgreSQL available at `127.0.0.1:8765`.


colinaumaty:
@mkysel Hi, this is a small fix in pkg/stress/watcher.go to stop the watcher goroutines when the subscription or log channel is closed, instead of continuing to receive from a closed channel. Would you mind taking a look when you have a chance? Thanks!

